PR #14, opened 2025-01-07
Model for running jags with stratifications.

sschildhauer:
Pull request for running a jags function with stratifications. Default values are set. Currently creates two outputs but will modify to create one major list of outputs.


d-morrison (changes requested):
Thanks for sending this to me; it's always ok to ask for help. Here are some quick notes:

[ ] documentation check: run `devtools::document()` in the R console (in the dcm project) and commit/push the updated .Rd files.
[ ] version check: the version in the PR needs to be one dev version (4th number) ahead of main branch before merging. Call usethis::use_version() in the R Console to increment the dev version number. This only needs to fixed right before merging the PR, so don't worry about it too much; I can always do it on my end if needed.
[ ] R-CMD-CHECK is throwing errors from other parts of the package; you can ignore until I tell you otherwise (basically, once main branch passes these checks, we want to keep it that way, but right now main also fails these)
[ ] lint-changed-files: if you look at https://github.com/UCD-SERG/dcm/pull/14/files there are annotations where the style issue are; please try to address them, and leave comments with `@d-morrison` on the lines where you're unsure what to do.

If you're still unsure what to do, please schedule a meeting with me and we'll take a look together.

d-morrison (approved):
Thanks!

PR #24, opened 2025-02-07
Adding attributes

sschildhauer:
I added model attributes as a return -- these are taken directly from the ggmcmc::ggs() function. These will show model inputs and will be used later for the diagnostic scripts. The diagnostic pull request that I will create shortly will not run until this  piece is merged into run_mod(), as the diagnostic scripts depend on having these attributes available. 

I also made some short edits: Took the reference out, small name changes in the documentation etc.

I am not totally sure why plot_jags_dens() is showing up in NAMESPACE as a diff, as I was working on that piece in another branch.


sschildhauer:
I am unsure why the spelling check is not running, when I run spelling::update_wordlist on my local it doesn't recommend any changes.

d-morrison (changes requested):
not sure the news.md PR # is correct, otherwise fine!

d-morrison (approved):
Looks good - thanks!

d-morrison:
> I am unsure why the spelling check is not running, when I run spelling::update_wordlist on my local it doesn't recommend any changes.

@sschildhauer I've noticed that sort of inconsistency too; eventually we should figure out why that is and maybe file a bug report on https://github.com/ropensci/spelling/issues. For now, let's just say that the github version is the gold standard, since our locals might differ from each other.

PR #27, opened 2025-02-07
Density dx function

sschildhauer:
Created a diagnostic function that takes an object from the run_mod() function and creates density plots. This uses the ggs_density function built into ggmcmc to create the figures. The default is to create a plot for every combination of iso/parameter, however you can specify if you want to only create figures for a specific iso or parameter -- for example, maybe you only want to compare the y1 across each iso. Furthermore, if you created a stratification in the run_mod() piece, you can also stratify your results here -- this piece might need a tiny bit of work and not totally sure if the output is the best.


R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,92 @@
...
+
+plot_jags_dens <- function(data,
+                           iso = unique(visualize_jags_sub$Iso_type),
+                           param = unique(visualize_jags_sub$Parameter_sub),
+                           strat = NA) {
+  visualize_jags <- data[["curve_params"]]
+  attributes_jags <- data[["attributes"]]
+
+  if (is.na(strat)) {
+    stratification <- c("No stratification")
```
d-morrison, line 51:
```suggestion
    stratification <- "No stratification"
```

will address the codefactor annotation below

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,92 @@
...
+  } else {
+    stratification <- unique(visualize_jags$Stratification)
```
d-morrison, line 53: we also need to create a unit test for this function, to make sure it continues to work as we expect it to as we make changes to the package over time. Please take a look at the [testing chapters](https://r-pkgs.org/testing-basics.html) of the R-packages book for an overview of how to construct unit tests, and take a look at the other tests that are now in the package (in the `tests/` subfolder). Glad to help you get the hang of this!

d-morrison (changes requested):
Looks good, but I'm stepping up the requirements for PRs going forward; now they will typically need to pass all checks marked as "required" in order to be approved. Please review the failing checks and my comments below - let me know if you have any questions, and glad to discuss over video if helpful.

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,92 @@
...
+    density_out <- list()
+    # Looping through the isos
+    for (j in iso) {
+      visualize_jags_plot <- visualize_jags_sub |>
+        dplyr::filter(.data$Iso_type == j)
+
+      # Will not loop through parameters, as we may want each to show on the
+      # same plot by default.
+      visualize_jags_plot <- visualize_jags_plot |>
+        dplry::filter(.data$Parameter_sub %in% param)
```
d-morrison, line 76:
```suggestion
        dplyr::filter(.data$Parameter_sub %in% param)
```

sschildhauer:
I attempted to review all of your existing changes. I merged all of your recent changes into this branch before pushing my changes.

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,89 @@
...
+#' @param param Specify [string] to produce plots of only a specific parameter,
+#' entered with quotes. Options include:
+#' - `alpha` = posterior estimate of decay rate
+#' - `r` = posterior estimate of shape parameter
+#' - `t1` = posterior estimate of time to peak
+#' - `y0` = posterior estimate of baseline antibody concentration
+#' - `y1` = posterior estimate of peak antibody concentration
+#' @param strat Specify [string] to produce plots of specific stratification
+#' entered in quotes.
+#' @return A [list] of [] objects producing density plots for all the specified
```
d-morrison, line 30: what kind of objects?

d-morrison (changes requested):
we'll chat today to discuss in more detail, but here are some preliminary notes; see also the errors in the github actions

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,89 @@
+
+#' @title Density Plot Diagnostics
+#' @author Sam Schildhauer
+#' @description
+#'  plot_jags_dens() takes a [list] output from the [serodynamics::run_mod()]
+#'  function to create density plots for each chain run in the mcmc estimation.
```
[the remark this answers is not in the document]
sschildhauer: Fixed recommended grammatical changes.

DESCRIPTION
```diff
@@ -32,9 +36,7 @@ Suggests:
     withr,
     vdiffr,
     fs
-Language: en-US
-Config/testthat/edition: 3
```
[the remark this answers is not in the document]
sschildhauer: Fixed discrepency

DESCRIPTION
```diff
@@ -23,6 +23,10 @@ Imports:
     tibble,
     utils
 Suggests: 
+    spelling,
+    testthat (>= 3.0.0)
```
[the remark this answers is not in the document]
sschildhauer: Fixed description

DESCRIPTION
```diff
@@ -23,6 +23,10 @@ Imports:
...
+Language: en-US
+Config/testthat/edition: 3
```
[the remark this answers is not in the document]
sschildhauer: Moved to correct location

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,89 @@
...
+#' - `y0` = posterior estimate of baseline antibody concentration
+#' - `y1` = posterior estimate of peak antibody concentration
+#' @param strat Specify [string] to produce plots of specific stratification
+#' entered in quotes.
+#' @return A [list] of [] objects producing density plots for all the specified
+#' input.
+#' @export
+#' @examples
+#' plot_jags_dens(
+#'     data = jags_out, #A [serodynamics::run_mod()] [list] output.
```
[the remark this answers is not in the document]
sschildhauer: Adapted from your and example from run_mod.

sschildhauer:
Hi Ezra, I went through the checks and got this to a point where I am a bit lost with the failures. It looks like the test is running fine. But, the might be a problem in \usage in the Rd file. I am going to open back up for review in the case where you have a quick solution for the issue. Thanks!

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,108 @@
...
+#'     iso = "hlya_IgG", #A [string] specifying antigen/antibody of interest.
+#'     param = "alpha",  #A [string] specifying parameter of interest.
+#'     strat = "strat")  #A [string] specifying stratification of interest.
+#'     }
+
+plot_jags_dens <- function(data,
+                           iso = unique(visualize_jags_sub$Iso_type),
+                           param = unique(visualize_jags_sub$Parameter_sub),
+                           strat = unique(visualize_jags$Stratification)) {
```
d-morrison, line 68: It's not a good idea to define the default inputs (`param` and `strat`) using a variable that is only defined inside the function body (`visualize_jags_sub`).

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,108 @@
...
+  visualize_jags <- data[["curve_params"]]
+  attributes_jags <- data[["attributes"]]
+
+  dens_strat_list <- list()
+  for (i in strat) {
```
d-morrison, line 73: in your example code, the `strat` input is defined as the string `"strat"`; is that what you meant to do?

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,108 @@
...
+#'  Defaults will produce every combination of antigen/antibody, parameters,
+#'  and stratifications, unless otherwise specified.
+#'  Antigen/antibody combinations and stratifications will vary by analysis.
+#'  The antibody dynamic curve includes the following parameters:
+#'  - y0 = baseline antibody concentration
+#'  - y1 = peak antibody concentration
+#'  - t1 = time to peak
+#'  - r = shape parameter
+#'  - alpha = decay rate
+#' @param name description
```
d-morrison, line 16: There isn't a `name` input in the actual function code below (lines 65-68); I think that's the cause of this error in R-CMD-check:

image

d-morrison (changes requested):
@sschildhauer This function appears to have some structural issues; please use `debug()` with your example code and step through the logic one line at a time to make sure the code is doing what you intended. Here are some guides on debugging:

- https://adv-r.hadley.nz/debugging.html
- https://d-rug.github.io/debugging_in_R/ (from the UCD R user group)

See inline comments for some hints!

R/plot_jags_densitydx.R
sschildhauer: Thanks for catching this. I specified one of the stratum run in the prior run_mod command.

R/plot_jags_densitydx.R
sschildhauer: I instead made this reference .data$, which I didn't originally know you could do. Thanks!

sschildhauer:
The last piece it looks like it is requesting is a reference, which I don't believe we have at this time. Any recommendations?

d-morrison (changes requested):
Good question - earlier today, I updated [_pkgdown.yml](https://github.com/UCD-SERG/serodynamics/blob/main/pkgdown/_pkgdown.yml) to define categories for the functions listed in the [References](https://ucd-serg.github.io/serodynamics/reference/index.html) page on the [package website](https://ucd-serg.github.io/serodynamics). 

`plot_jags_dens` now needs to be added to [_pkgdown.yml](https://github.com/UCD-SERG/serodynamics/blob/main/pkgdown/_pkgdown.yml).

See https://pkgdown.r-lib.org/articles/pkgdown.html#reference for more explanation of how `_pkgdown.yml` works.

sschildhauer:
I think it finally might be there!

tests/testthat/test-plot_jags_densitydx.R
```diff
@@ -0,0 +1,41 @@
...
+        results <- run_mod(
+          data = dataset, #The data set input
+          file_mod = fs::path_package("serodynamics", "extdata/model.jags"),
+          nchain = 2, #Number of mcmc chains to run
+          nadapt = 100, #Number of adaptations to run
+          nburn = 100, #Number of unrecorded samples before sampling begins
+          nmc = 100,
+          niter = 100, #Number of iterations
+          strat = "strat"
+        )  |>
```
d-morrison, line 29: Pre-save the results up to here as [package data](https://r-pkgs.org/data.html#sec-data-data) to make this test run faster?

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,110 @@
...
+#'  Defaults will produce every combination of antigen/antibody, parameters,
+#'  and stratifications, unless otherwise specified.
+#'  Antigen/antibody combinations and stratifications will vary by analysis.
+#'  The antibody dynamic curve includes the following parameters:
+#'  - y0 = baseline antibody concentration
+#'  - y1 = peak antibody concentration
+#'  - t1 = time to peak
+#'  - r = shape parameter
+#'  - alpha = decay rate
+#' @param data A [base::list()] outputted from run_mod().
```
d-morrison, line 16:
```suggestion
#' @param data A [list] outputted from run_mod().
```

Crosslinks to functions and classes from the 6 default R packages ("datasets"  "utils"     "grDevices" "graphics"  "stats"     "methods") don't need to be explicitly prefixed with their package name; `roxygen2` will figure it out for you.

Also, when you're referring to a function, I think it's best to include the parentheses in the crosslink, but when you're referring to an object class, like you're doing here, I think it's best to omit the parentheses.

pkgdown/_pkgdown.yml
```diff
@@ -19,6 +19,9 @@ reference:
   - load_data
   - as_case_data
   - prep_data
+- title: Model diagnostics
+  contents:
+  - plot_jags_dens
```
d-morrison, line 24: Let's move this section after the "Model seroresponse" section, since we use it with the output of `run_mod()`

tests/testthat/test-plot_jags_densitydx.R
```diff
@@ -0,0 +1,41 @@
...
+    set.seed(1)
+    library(dplyr)
+    strat1 <- serocalculator::typhoid_curves_nostrat_100 |>
+      sim_case_data(n = 100) |>
+      mutate(strat = "stratum 2")
+    strat2 <- serocalculator::typhoid_curves_nostrat_100 |>
+      sim_case_data(n = 100) |>
+      mutate(strat = "stratum 1")
+
+    dataset <- bind_rows(strat1, strat2)
```
d-morrison, line 15: replace this simulated data with the `nepal_sees` data?

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,110 @@
...
+#'   set.seed(1)
+#'   library(dplyr)
+#'   strat1 <- serocalculator::typhoid_curves_nostrat_100 |>
+#'     sim_case_data(n = 100) |>
+#'     mutate(strat = "stratum 2")
+#'   strat2 <- serocalculator::typhoid_curves_nostrat_100 |>
+#'     sim_case_data(n = 100) |>
+#'     mutate(strat = "stratum 1")
+#'
+#'   Dataset <- bind_rows(strat1, strat2)
```
d-morrison, line 45: let's replace this simulated data with the `nepal_sees` data?

R/plot_jags_densitydx.R
```diff
@@ -0,0 +1,110 @@
...
+#'   jags_out <- run_mod(
+#'     data = Dataset, # The data set input
+#'     file_mod = fs::path_package("serodynamics", "extdata/model.jags"),
+#'     nchain = 2, # Number of mcmc chains to run
+#'     nadapt = 100, # Number of adaptations to run
+#'     nburn = 100, # Number of unrecorded samples before sampling begins
+#'     nmc = 500,
+#'     niter = 1000, # Number of iterations
+#'     strat = "strat"
+#'   ) # Variable to be stratified
```
d-morrison, line 56: Pre-save the results up to here as [package data](https://r-pkgs.org/data.html#sec-data-data) to make this example shorter and faster?

d-morrison (changes requested):
Looking good! I'm requesting a few more changes to fine-tune things bit more.

sschildhauer:
I am not totally sure why this last piece isn't passing. It looks like I need to specify the columns types, which I am not totally sure how to go about when I am loading the object from the data folder. Thank you for taking a look!

tests/testthat/test-plot_jags_densitydx.R
```diff
@@ -0,0 +1,25 @@
+
+test_that(
+  desc = "results are consistent with ggplot output",
+  code = {
+    library(runjags)
```
d-morrison, line 5:
```suggestion
    skip_if(getRversion() < "4.4.1") # 4.3.3 had issues
    library(runjags)
```
There are some situations where R version 4.3 produces slightly different ggplot outputs from R 4.4, which appears to be causing a snapshot discrepancy issue when we run R-CMD-check on "ubuntu-oldrel" (oldrel = R v4.3.3 at the moment). I downloaded the snapshot from that check and compared it with the snapshot from R 4.4, and the differences seem insignificant:

image

So, I think we can just skip this check on R 4.3 (see suggested edit above)

R/plot_jags_densitydx.R
```diff
@@ -33,34 +33,18 @@
...
 #'   library(ggmcmc)
-#'   set.seed(1)
 #'   library(dplyr)
-#'   strat1 <- serocalculator::typhoid_curves_nostrat_100 |>
-#'     sim_case_data(n = 100) |>
-#'     mutate(strat = "stratum 2")
-#'   strat2 <- serocalculator::typhoid_curves_nostrat_100 |>
-#'     sim_case_data(n = 100) |>
-#'     mutate(strat = "stratum 1")
+#'   set.seed(1)
```
d-morrison, line 37: do we need to `set.seed()` now that we are using pre-run results?
```suggestion
```

tests/testthat/_snaps/plot_jags_densitydx/typhoid-plot.svg
d-morrison, line 1: these graphs look good!

d-morrison (changes requested):
Thanks for adding the snapshot! See possible solution in comments

tests/testthat/test-plot_jags_densitydx.R
sschildhauer: Okay, thanks for the tip!

d-morrison (approved):
looks good to me - thanks!

PR #41, opened 2025-02-17
Adding SEES data to data folder

sschildhauer:
Adding SEES serokinetics data from nepal to the data folder to be accessible to those using the package, as well as to be used in examples and tests.


R/Nepal_SEES.R
```diff
@@ -0,0 +1,23 @@
...
+#'  \item{bldculres}{Pathogen participant tested positive for; Typhoid or
+#'  paratyphoid}
+#'  \item{antigen_iso}{The antigen/antibody combination included in the assay}
+#'  \item{studyvisit}{Categorical estimated time frame for when sample was
+#'  taken; 28 days, 3_months, 6_months, 12_months, baseline, or 18_months}
+#'  \item{dayssincefeveronset}{Continuous measurement showing how exact days
+#'  since symptom onset}
+#'  \item{result}{Continuous variable describing ELISA result}
+#' }
+#' @source Link to SEES study?
```
d-morrison, line 23: You just need one more line here:
```suggestion
#' @source Link to SEES study?
"nepal_sees"
```
see https://r-pkgs.org/data.html#sec-documenting-data

d-morrison (changes requested):
almost there I think!

DESCRIPTION
```diff
@@ -46,4 +46,7 @@ Config/needs/test: withr, vdiffr, here, readr
 Config/needs/check: commonmark, xml2
 Remotes: 
     ucd-serg/serocalculator
+Depends: 
+    R (>= 3.5)
```
d-morrison, line 50:
```suggestion
    R (>= 4.1.0)
```
We actually need R ≥ 4.1.0 to use the `|>` notation

sschildhauer: Changed version

R/Nepal_SEES.R
sschildhauer: Added "nepal_sees" to document.

R/Nepal_SEES.R
```diff
@@ -0,0 +1,24 @@
...
+#' "nepal_sees"
```
d-morrison, line 24: this line shouldn't be commented out:
```suggestion
"nepal_sees"
```

d-morrison (commented):
the format for documenting the data is rigid; please see suggested edit

sschildhauer:
It looks like the main piece that still isn't passing is the @source link to the data, and I don't currently have a URL link. Would this URL link usually be a link to the actual data set, or a link to a description of the data and where it came from? Thanks!

R/Nepal_SEES.R
```diff
@@ -0,0 +1,24 @@
+#'
+#'
+#' SEES Typhoid data
+#'
+#' A subset of data from the SEES project highlighting Typhoid longitudinal
+#' data from Nepal.
+#'
+#' @format ## `nepal_sees`
+#' A [base:data.frame()] with 904 rows and 8 columns:
```
d-morrison, line 9: I think this change will fix 
image


```suggestion
#' A [base::data.frame()] with 904 rows and 8 columns:
```

data-raw/nepal_sees.R
d-morrison, line 1: `data-raw` scripts are supposed to show *how* you created the `nepal_sees` dataset, starting with reading a raw csv file (or whatever format you received it) and including all `filter()`, `mutate()`, or other preprocessing steps in between.

Examples:

- https://github.com/tidyverse/nycflights13/blob/main/data-raw/airports.R
- https://github.com/UCD-SERG/serocalculator/blob/main/data-raw/sees_pop_data_pakistan_100.R

d-morrison (changes requested):
Making progress! Please see specific comments.

data-raw/nepal_sees.R
```diff
@@ -0,0 +1,3 @@
+## code to prepare `nepal_sees` dataset goes here
```
d-morrison, line 1: (you're supposed to replace this line with your preprocessing code)

R/Nepal_SEES.R
sschildhauer: Fixed issue

PR #63, opened 2025-02-28
nepal_sees and jags_out

sschildhauer:
Fixing nepal_sees in data folder and creating new R object for nepal_sees_jags_out. nepal_sees R object was only two lines of code, as something went wrong with the import. Also added documentation for new jags_out object using nepal_sees in run_mod.


sschildhauer:
I am correcting the nepal_sees data, which had uploaded in a strange format -- it had created an R object of the title of the csv and not the actual content. 
I am also creating an object for jags.post output from the nepal_sees data. All of the jags output are pretty big, so I applied 'compress = "xz" ' in the use_data function. This changed the object from 5MB to 3MB. However, it looks like the size might still be creating some issues (there is a note). I am interested to hear your advice on dealing with this issue. 
I am planning on using this jags_post object as example and test for all of the diagnostic functions.
Thank you!

R/nepal_sees_jags_post.R
```diff
@@ -0,0 +1,19 @@
...
+#' stratifying by bldculres, which is the diagnosis type (typhoid or
+#' paratyphoid).
+#'
+#' @format ## `nepal_sees_jags_post`
+#' A jags.post [list] object or multiple jags.post [list]
+#' if stratified. Returned as a [list] of class [runjags::runjags-class]
+#' \describe{
+#'  \item{curve_params}{A [data.frame] titled `curve_params` that contains the
+#'   posterior distribution}
+#'  \item{jags.post}{An object with class [runjags::runjags]}
```
d-morrison, line 15:
```suggestion
#'  \item{jags.post}{A [list] of [runjags::runjags-class] objects}
```
if I'm understanding correctly:
image

`[runjags::runjags]` unfortunately redirects to ``?runjags::`runjags-package` `` instead of ``?runjags::`runjags-class` ``

Fortunately, `{roxygen2}` will format the resulting Rd doc file nicely:

image

R/nepal_sees_jags_post.R
```diff
@@ -0,0 +1,19 @@
...
+#'
+#' SEES Typhoid run_mod jags output
+#'
+#' An [serodynamics::run_mod()] output using sees_nepal data as input, 
+#' stratifying by bldculres, which is the diagnosis type (typhoid or
+#' paratyphoid).
+#'
+#' @format ## `nepal_sees_jags_post`
+#' A jags.post [list] object or multiple jags.post [list]
+#' if stratified. Returned as a [list] of class [runjags::runjags-class]
```
d-morrison, line 11:
```suggestion
#' A [list] consisting of the following named elements:
```

data/nepal_sees_jags_post.rda
d-morrison, line 1: This file is 2.99 MB, which is typically [too large for CRAN](https://r-pkgs.org/data.html#sec-data-data-raw:~:text=updated%20as%20necessary.-,Submitting%20to%20CRAN,-Generally%2C%20package%20data):

image

The file size is triggering a Note in R-CMD-check for the current R version (4.4.2):
image

Interestingly, the in-development version of R downgraded this issue from a "Note" to an "Info":

image

So it's possible CRAN's size limit will be increased soon. For now though, let's reduce this file size as much as we can:

* reduce the number of MCMC samples to something smaller, by increasing the thinning parameter.
* most of the size comes from the list of `jags.post` objects; do we even need to keep those in the output of `run_mod()`?
* If we do need to keep them in general, can we remove them post-hoc just for this example object? In other words, can we set `nepal_sees_jags_post$jags.post <- NULL` in a preprocessing step before we call `use_data()`? I'm not sure whether we will need the `jags.post` component for any of the examples or tests we are going to want.

d-morrison (commented):
Thanks for this PR! Looks good overall, but we have to figure out what to do about the file size of `nepal_sees_jags_post.rda`; see comments.

R/nepal_sees_jags_post.R
d-morrison: (meant to formally "request changes" but clicked "comment" instead in the review menu.

data/nepal_sees_jags_post.rda
sschildhauer: I decided to go with your third recommendation and take out jags.post. But, I don't think we can take out the raw jags.post from the run_mod output. As it is now, the curve params file does not contain model parameters for each individual, it only contains the information for the "new person". So, for example, the diagnostic code that Kwan is working on might not be possible if we take out the raw piece. Happy to discuss further.

d-morrison: @sschildhauer good point; maybe we could keep just the mcmc samples for the other IDs instead of the full jags.post output. I think we can hold off on deciding that until we get further along in developing the set of diagnostic functions.

PR #64, opened 2025-03-05
Traceplot for diagnostics

sschildhauer:
This function takes a jags_post output (from run_mod) list and creates traceplots with the possibility of creating plots for all combinations of isotype, antigen, parameter, or stratification. This function works in a very similar manner to plot_jags_dens.


R/plot_jags_tracedx.R
```diff
@@ -0,0 +1,93 @@
...
+#'  Defaults will produce every combination of antigen/antibody, parameters,
+#'  and stratifications, unless otherwise specified.
+#'  Antigen/antibody combinations and stratifications will vary by analysis.
+#'  The antibody dynamic curve includes the following parameters:
+#'  - y0 = baseline antibody concentration
+#'  - y1 = peak antibody concentration
+#'  - t1 = time to peak
+#'  - r = shape parameter
+#'  - alpha = decay rate
+#' @param data A [list] outputted from run_mod().
```
d-morrison, line 16:
```suggestion
#' @param data A [list] outputted from [run_mod()].
```

R/plot_jags_tracedx.R
```diff
@@ -0,0 +1,93 @@
...
+#' @param param Specify [character] string to produce plots of only a
+#' specific parameter, entered with quotes. Options include:
+#' - `alpha` = posterior estimate of decay rate
+#' - `r` = posterior estimate of shape parameter
+#' - `t1` = posterior estimate of time to peak
+#' - `y0` = posterior estimate of baseline antibody concentration
+#' - `y1` = posterior estimate of peak antibody concentration
+#' @param strat Specify [character] string to produce plots of specific
+#' stratification entered in quotes.
+#' @return A [base::list()] of [ggplot2::ggplot()] objects producing trace
```
d-morrison, line 29:
```suggestion
#' @return A [list] of [ggplot2::ggplot] objects producing trace
```

tests/testthat/_snaps/plot_jags_tracedx/tracedx-typhoid-plot.svg
d-morrison, line 1: move legend to bottom?

tests/testthat/test-plot_jags_tracedx.R
```diff
@@ -0,0 +1,25 @@
...
+  desc = "results are consistent with traceplot ggplot output",
+  code = {
+    skip_if(getRversion() < "4.4.1") # 4.3.3 had issues
+    library(runjags)
+    library(dplyr)
+
+    data <- serodynamics::nepal_sees_jags_post |>
+      suppressWarnings()
+
+    withr::with_seed(
```
d-morrison, line 12: is with_seed() necessary here?

d-morrison (changes requested):
Looks good! A few suggestions and questions

tests/testthat/_snaps/plot_jags_tracedx/tracedx-typhoid-plot.svg
d-morrison: (in function)

tests/testthat/test-plot_jags_tracedx.R
sschildhauer: Hmmm true, I don't think it is considering the run_mod has already been run. I will take this piece out.

sschildhauer:
Integrated recommendations. Moved example to its own file. Took out unnecessary code from example and test. I believe ready for another review now that passing all checks.

R/plot_jags_tracedx.R
```diff
@@ -37,14 +37,12 @@
 #'
 #'   data <- serodynamics::nepal_sees_jags_post
```
d-morrison, line 38: formatting is off here; move examples to separate file?
```suggestion
#' data <- serodynamics::nepal_sees_jags_post
```

R/plot_jags_tracedx.R
```diff
@@ -0,0 +1,92 @@
...
+#'
+#'   data <- serodynamics::nepal_sees_jags_post
+#'
+#'# Specifying isotype, parameter, and stratification for traceplot.
+#' plot_jags_trace(
+#'     data = data,
+#'     iso = "HlyE_IgA",
+#'     param = "alpha",
+#'     strat = "typhi")
+#'     }
```
d-morrison, line 46: In the [rendered example](https://ucd-serg.github.io/serodynamics/preview/pr64/reference/plot_jags_trace.html), there's a bit of maybe-unnecessary text output:

image

```
#> $typhi
#> $typhi$HlyE_IgA
```

Is this intentional? If not, can we get rid of it?

d-morrison (changes requested):
Looks good! I left a couple more polishing notes. @Kwan-Jenny please also review.

R/plot_jags_tracedx.R
d-morrison: on second look, the indentation here is correct; it's the indentation later in the example that seems off. Either way, if we move the example to an external file the linter will check it automatically and https://github.com/r-lib/styler or https://github.com/posit-dev/air can fix it for you.

sschildhauer: Moved example to its own file

R/plot_jags_tracedx.R
sschildhauer: I am unsure how to get rid of this. The object that is being created is a list of ggplot objects for every antigen/iso combination. Any tips on how to not output this in the console would be appreciated -- I was having a hard time finding solutions online.

d-morrison: > I am unsure how to get rid of this. The object that is being created is a list of ggplot objects for every antigen/iso combination. Any tips on how to not output this in the console would be appreciated -- I was having a hard time finding solutions online.

(see discussion in teams.)

tests/testthat/_snaps/plot_jags_tracedx/tracedx-typhoid-plot.svg
d-morrison, line 1: - [ ] The x axis label ("log(value)") doesn't seem correct here; aren't those iterations?

- [ ] As discussed in teams, let's switch to using logarithmic *spacing* (e.g., `scale_y_log10()`) instead of logarithmic *scaling*

d-morrison (changes requested):
as in the other PRs, I'd like to tidy up the list printing; see also inline comments in https://github.com/UCD-SERG/serodynamics/pull/64/files.

R/plot_jags_tracedx.R
sschildhauer: I was able to get rid of text when there is a single plot. However, when there are more plots the text will still appear.

tests/testthat/_snaps/plot_jags_tracedx/tracedx-typhoid-plot.svg
sschildhauer: Thank you for double checking these. Both should be fixed.

R/plot_jags_tracedx.R
```diff
@@ -60,19 +60,23 @@ plot_jags_trace <- function(data,
...
+        ggplot2::labs(x = "iterations", y = "log(value)") +
+        ggplot2::theme(legend.position = "bottom") +
+        ggplot2::scale_y_log10()
       trace_out[[j]] <- traceplot
     }
     trace_strat_list[[i]] <- trace_out
   }
+  #Printing only one plot if only one exists.
+  if (length(trace_strat_list) == 1) {
+    trace_strat_list <- trace_strat_list[[1]][[iso]]
```
d-morrison, line 79: what happens if there's one stratum (so that `length(trace_strat_list) == 1`) but multiple `iso`s? I suspect this line as currently written will throw an error; please add a unit test for this scenario and fix if needed?

inst/examples/examples-plot_jags_tracedx.R
```diff
@@ -0,0 +1,12 @@
+
+if (!is.element(runjags::findjags(), c("", NULL))) {
```
d-morrison, line 2: now that we have pre-saved `serodynamics::nepal_sees_jags_post`, can we remove this `if ()` condition? I think this example will still be runnable if jags isn't installed.

R/plot_jags_tracedx.R
```diff
@@ -60,19 +60,23 @@ plot_jags_trace <- function(data,
...
       attributes(visualize_jags_plot) <- c(attributes(visualize_jags_plot),
                                            attributes_jags)
       # Creating density plot
       traceplot <- ggmcmc::ggs_traceplot(visualize_jags_plot) +
         ggplot2::theme_bw() +
-        ggplot2::labs(x = "log(value)") +
-        ggplot2::theme(legend.position = "bottom")
+        ggplot2::labs(x = "iterations", y = "log(value)") +
+        ggplot2::theme(legend.position = "bottom") +
+        ggplot2::scale_y_log10()
```
d-morrison, line 72: Try this:

```suggestion
        ggplot2::scale_y_log10(labels = scales::label_comma())
```

It should improve the formatting of the y-axis tick labels for the `alpha` parameter

R/plot_jags_tracedx.R
```diff
@@ -60,19 +60,23 @@ plot_jags_trace <- function(data,
...
+                                         i))
       # Assigning attributes, which are needed to run ggs_density
       attributes(visualize_jags_plot) <- c(attributes(visualize_jags_plot),
                                            attributes_jags)
       # Creating density plot
       traceplot <- ggmcmc::ggs_traceplot(visualize_jags_plot) +
         ggplot2::theme_bw() +
-        ggplot2::labs(x = "log(value)") +
-        ggplot2::theme(legend.position = "bottom")
+        ggplot2::labs(x = "iterations", y = "log(value)") +
```
d-morrison, line 70: The y axis tick marks are showing the parameter values on their original scale now, so I think we don't want `"log(value)"`, just `"value"`, or maybe `"parameter value"`)
```suggestion
        ggplot2::labs(x = "iterations", y = "parameter value") +
```

d-morrison (changes requested):
Good progress! A few more notes in https://github.com/UCD-SERG/serodynamics/pull/64/files/

inst/examples/examples-plot_jags_tracedx.R
sschildhauer: Took this out.

R/plot_jags_tracedx.R
sschildhauer: Fixed to look at if sum(lengths()) == 1.. This should work. Thanks for catching this!

d-morrison (approved):
In the pkgdown website, the example figure for `plot_jags_trace()` gets a little squashed, so this is probably one case where we *don't* need to show the figure for all parameters:

image

I'm not sure though, so I'll merge as-is, and we can fix it later if we want.

PR #66, opened 2025-03-06
Effective sample size diagnostics

sschildhauer:
This function creates a plot that shows the proportion of effective samples drawn from the total samples drawn for each estimated parameter. This function uses ggs_effective to create the plot.


R/plot_jags_effectivedx.R
```diff
@@ -0,0 +1,95 @@
...
+#'
+#' plot_jags_effect(
+#'     data = data, #A [serodynamics::run_mod()] [list] output.
+#'     iso = "HlyE_IgA", #A [character] string specifying
+#'     #nantigen/antibody of interest.
+#'     param = "alpha",  #A [character] string specifying parameter of
+#'     # interest.
+#'     strat = "typhi")  #A [character] string specifying
+#'     # stratification of interest.
+#'     }
```
d-morrison, line 51: The indentation of the closing `}` looks off. 
Let's switch to using separate example files for any examples that are more than ~5 lines: https://roxygen2.r-lib.org/articles/rd.html?q=examples#examples:~:text=put%20them%20in-,separate%20files,-and%20use%20%40example

Then the linter should catch these kinds of indentation issues automatically.

R/plot_jags_effectivedx.R
```diff
@@ -0,0 +1,95 @@
...
+#' @param param Specify [character] string to produce plots of only a
+#' specific parameter, entered with quotes. Options include:
+#' - `alpha` = posterior estimate of decay rate
+#' - `r` = posterior estimate of shape parameter
+#' - `t1` = posterior estimate of time to peak
+#' - `y0` = posterior estimate of baseline antibody concentration
+#' - `y1` = posterior estimate of peak antibody concentration
+#' @param strat Specify [character] string to produce plots of specific
+#' stratification entered in quotes.
+#' @return A [base::list()] of [ggplot2::ggplot()] objects showing the
```
d-morrison, line 30: see comments on previous PRs; let's not use parentheses when cross-linking object class names.

d-morrison (changes requested):
Looks good! I don't have time for a complete line-by-line review right now, but I caught a few minor formatting issues. Please address those, and then request a review from @Kwan-Jenny; I can do another pass after Kwan approves.

R/plot_jags_effectivedx.R
```diff
@@ -0,0 +1,95 @@
...
+#' @examples
+#' if (!is.element(runjags::findjags(), c("", NULL))) {
+#'   library(runjags)
+#'   library(ggmcmc)
+#'   library(dplyr)
+#'
+#'   data <- serodynamics::nepal_sees_jags_post
+#'
+#' plot_jags_effect(
+#'     data = data, #A [serodynamics::run_mod()] [list] output.
```
d-morrison, line 44: @sschildhauer one more note; in the examples sections, the `[]` syntax doesn't do anything, so let's not use it; take a look at https://ucd-serg.github.io/serodynamics/preview/pr66/reference/plot_jags_effect.html to see how the example gets rendered.
Instead, you can either just use backticks (for example \`list\`), quotes, or just plain text. 
In fact, this amount of commenting is pretty unusual for examples; take a look at the tidyverse packages to see how examples are usually formatted

d-morrison (commented):
(added one more note)

R/plot_jags_effectivedx.R
sschildhauer: Moved example to separate folder.

R/plot_jags_effectivedx.R
sschildhauer: Took parentheses out.

sschildhauer:
I fixed requested comments by Ezra  and moved example to inst/examples/. I am now getting an error that I don't quite understand. It looks like it is an error in n() from dplyr::summarize... but I am confused as to why an error would now be triggered from dplyr. I didn't really change much that would seem to set off a trigger and examples are running fine when I conduct them internally with addins.

Kwan-Jenny (dismissed):

d-morrison:
> I fixed requested comments by Ezra and moved example to inst/examples/. I am now getting an error that I don't quite understand. It looks like it is an error in n() from dplyr::summarize... but I am confused as to why an error would now be triggered from dplyr. I didn't really change much that would seem to set off a trigger and examples are running fine when I conduct them internally with addins.

@sschildhauer @Kwan-Jenny Sorry to jump in unrequested; the error you're facing is a tricky one that I'm partly to blame for, so I want to help fix it. The error is actually coming from `ggmcmc::ggs_effective()`, which I can tell from the `Backtrace` part of the error in https://github.com/UCD-SERG/serodynamics/actions/runs/13715802955/job/38360355850?pr=66:

image

The problem is that `ggmcmc` [doesn't import](https://github.com/xfim/ggmcmc/blob/352a0f8139a023cc30db9f18477419f8cf4f7623/R/help.R#L12) the function `n()` from `dplyr`, and attempts to use it [here](https://github.com/xfim/ggmcmc/blob/352a0f8139a023cc30db9f18477419f8cf4f7623/R/ggs_effective.R#L33)

That's normally not a problem for ggmcmc because it [`Depends`](https://r-pkgs.org/dependencies-in-practice.html#sec-dependencies-in-depends) on `dplyr` ([here](https://github.com/xfim/ggmcmc/blob/352a0f8139a023cc30db9f18477419f8cf4f7623/DESCRIPTION#L14)).
But if you never call `library(ggmcmc)`, then `dplyr` never gets loaded, so `n()` isn't defined for use in `ggmcmc::ggs_effective()`.
You're not encountering this problem on your local computers because I unwisely set up dplyr to get automatically loaded when R starts in an interactive session, using the [`.Rprofile` file](https://github.com/UCD-SERG/serodynamics/blob/250f1a612582447451a74da3f1d2baba3da04809/.Rprofile#L11) in the serodynamics package. But that .Rprofile code doesn't get run on Github, so `n()` never gets defined.

In the long-term, we should ask the ggmcmc authors to import `n()` from `dplyr` (by filing an [Issue](https://github.com/xfim/ggmcmc/blob/master/R/help.R#L12) with a [reprex](https://reprex.tidyverse.org/)). 

But in the short term, we can fix this problem for ourselves by importing `n()` into `serodynamics`, with `usethis::use_import_from("dplyr", "n")`. 

We should also edit `.Rprofile` to not auto-load `dplyr`; I'll do that now.

sschildhauer:
Thank you for reviewing Ezra, this was a tough one. I used use_import_from to add n().. unsure why it still is not passing.

d-morrison:
> Thank you for reviewing Ezra, this was a tough one. I used use_import_from to add n().. unsure why it still is not passing.

@sschildhauer @Kwan-Jenny I'm not sure why either, but I noticed that ggmcmc actually fixed this issue on their end in their most recent commit (https://github.com/xfim/ggmcmc/commit/352a0f8139a023cc30db9f18477419f8cf4f7623, May 12, 2023). They haven't updated their [CRAN ](https://cran.r-project.org/web/packages/ggmcmc/index.html) version since 2021-02-10. I pinged them about that [here](https://github.com/xfim/ggmcmc/issues/82); hopefully they'll update the CRAN version soon, but in the meantime we can use the github version.

I added commit https://github.com/UCD-SERG/serodynamics/pull/66/commits/12e151f43589d843a1c96b452aedae40fc10be93 to this branch; it modifies `DESCRIPTION` accordingly; on your local computers you should run `devtools::install_github("xfim/ggmcmc")` to make sure you have the latest version of ggmcmc

R/plot_jags_effectivedx.R
```diff
@@ -0,0 +1,83 @@
...
+      # Creating density plot
+      eff <- ggmcmc::ggs_effective(visualize_jags_plot) +
+        theme_bw()
+      eff_out[[j]] <- eff
+    }
+    eff_strat_list[[i]] <- eff_out
+  }
+  #Printing only one plot if only one exists.
+  if (length(eff_strat_list) == 1) {
+    eff_strat_list <- eff_strat_list[[1]][[iso]]
```
d-morrison, line 80: same concern as in the other PRs; please test what happens with one stratum, multiple isos

inst/examples/examples-plot_jags_effectivedx.R
```diff
@@ -0,0 +1,10 @@
+
+if (!is.element(runjags::findjags(), c("", NULL))) {
+
+  data <- serodynamics::nepal_sees_jags_post
+
+  plot_jags_effect(data = data,
+                   iso = "HlyE_IgA",
+                   param = "alpha",
```
d-morrison, line 8: maybe plot all params like in the unit test; the example plot only has one plotted point right now, which doesn't really illustrate what this function is for:
https://ucd-serg.github.io/serodynamics/preview/pr66/reference/plot_jags_effect.html#ref-examples

d-morrison (changes requested):
coming along - see comments.

inst/examples/examples-plot_jags_effectivedx.R
```diff
@@ -0,0 +1,10 @@
+
+if (!is.element(runjags::findjags(), c("", NULL))) {
```
d-morrison, line 2: can we delete this if() condition?

d-morrison (approved):
Looks good - thanks!

PR #67, opened 2025-03-06
Diagnostic Rhat function

sschildhauer:
Creating a diagnostic Rhat (Potential Scale Reduction Factor) function, which is a good indicator of convergence. This provides a numerical index of convergence rather than just observing plots. This looks at the ratio of the between and within chain variation. With convergence Rhat should be 1, but many consider a Rhat < 1.10 acceptable. This function can be subset for antigen/iso, parameter, or stratification.


sschildhauer:
One lint review is not passing because I titled the function plot_jags_Rhat, with a capitol R. However, I think the capitol R is appropriate here as that is how the metric is listed in the literature.

d-morrison (changes requested):
Looking good! A few notes. @Kwan-Jenny, please also review.

inst/examples/examples-plot_jags_rhatdx.R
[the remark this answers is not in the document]
sschildhauer: Yes, I was subsetting down to a single antigen/isotype/parameter combination. I changed the example to output all parameters for one antigen/isotype.

d-morrison (changes requested):
as in the other PRs, I'd like to tidy up the list printing; otherwise, looks good to me; let's have @Kwan-Jenny and @kaiemjoy review it before merging, to make sure the functionality and documentation work for them.

R/plot_jags_rhatdx.R
```diff
@@ -61,19 +61,25 @@ plot_jags_Rhat <- function(data,  # nolint: object_name_linter
 
       visualize_jags_plot <- visualize_jags_plot |>
         # Changing parameter name to reflect the input
-        dplyr::mutate(Parameter = paste0("iso = ", j, ", parameter = ",
-                                         .data$Parameter_sub, ", strat = ",
-                                         i),
+        dplyr::mutate(Parameter = paste0(.data$Parameter_sub),
```
d-morrison, line 64: this line seems like it can be simplified, or possibly eliminated?

```suggestion
        dplyr::mutate(Parameter = .data$Parameter_sub,
```

d-morrison (dismissed):
Looks good! I marked one line that you might be able to simplify.

R/plot_jags_rhatdx.R
d-morrison: > this line seems like it can be simplified, or possibly eliminated?

@sschildhauer up to you if you want to try to change this or not; otherwise, you should be able to merge as-is (after resolving the merge conflict

sschildhauer:
Hi Ezra, it looks like merging is blocked for me.

d-morrison (approved):
Looks good!

d-morrison:
> Hi Ezra, it looks like merging is blocked for me.

@sschildhauer ah; looks like we have the repo set to dismiss old approvals when new commits get added.

PR #71, opened 2025-03-11
Updating density diagnostic function

sschildhauer:
Updating density dx function. Moving example to inst/.  Adding piece to only call single plot.  Adding scale_x_10 to plot. Updating test snap.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -38,12 +38,12 @@ jobs:
           del wjags.exe
         shell: cmd
       
+      # from https://github.com/ku-awdc/runjags/issues/2:
       - name: Install JAGS (macOS)
         if: runner.os == 'macOS'
         run: |
-          brew install pkg-config
-          brew install jags
-          brew upgrade
+          sudo curl --location https://sourceforge.net/projects/mcmc-jags/files/JAGS/4.x/Mac%20OS%20X/JAGS-4.3.1.pkg -o /usr/local/JAGS-4.3.1.pkg
+          sudo installer -pkg /usr/local/JAGS-4.3.1.pkg -target /usr/local/bin/
 
       - name: Install JAGS (Ubuntu)
         if: runner.os == 'Linux'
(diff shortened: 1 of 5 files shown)


R/plot_jags_densitydx.R
```diff
@@ -26,26 +26,10 @@
...
-#' plot_jags_dens(
-#'     data = data, #A [serodynamics::run_mod()] [list] output.
-#'     iso = "HlyE_IgA", #A [character] string specifying
-#'     #nantigen/antibody of interest.
-#'     param = "alpha",  #A [character] string specifying parameter of
-#'     # interest.
-#'     strat = "typhi")  #A [character] string specifying
-#'     # stratification of interest.
-#'     }
+#' @examples inst/examples/examples-plot_jags_densitydx.R
```
d-morrison, line 32:
```suggestion
#' @example inst/examples/examples-plot_jags_densitydx.R
```
This is a frustrating convention from roxygen2: `@examples` is for inline examples, `@example` (no `s`) is for linking to external example files

inst/examples/examples-plot_jags_densitydx.R
```diff
@@ -0,0 +1,11 @@
+
+if (!is.element(runjags::findjags(), c("", NULL))) {
```
d-morrison, line 2: (as in the other PR, I think we can remove this line and the matching `}`, now that we have pre-saved `serodynamics::nepal_sees_jags_post`)

R/plot_jags_densitydx.R
```diff
@@ -76,18 +60,23 @@ plot_jags_dens <- function(data,
...
-                      value = log(.data$value))
+                                         i))
       # Assigning attributes, which are needed to run ggs_density
       attributes(visualize_jags_plot) <- c(attributes(visualize_jags_plot),
                                            attributes_jags)
       # Creating density plot
       densplot <- ggmcmc::ggs_density(visualize_jags_plot) +
         ggplot2::theme_bw() +
-        ggplot2::labs(x = "log(value)")
+        ggplot2::labs(x = "log(value)") +
```
d-morrison, line 70:
```suggestion
        ggplot2::labs(x = "parameter value") +
```
(or leave it as whatever `ggs_density()` defaults to?)

R/plot_jags_densitydx.R
```diff
@@ -76,18 +60,23 @@ plot_jags_dens <- function(data,
...
+        ggplot2::scale_x_log10()
```
d-morrison, line 71:
```suggestion
        ggplot2::scale_x_log10(labels = scales::label_comma())
```

d-morrison (changes requested):
looking good so far; see comments

sschildhauer:
This is a pull request from a while back with some small changes to the density diagnostic function. Primarily adding the scale_x_10 to the plot.

sschildhauer:
It looks like these fixes were actually completed in a separate PR. Okay if I go ahead and pr_finish? @d-morrison

PR #76, opened 2025-03-20
Add class attribute to runmod output

sschildhauer:
Assigning a class of sr_mod_list to the output of run_mod.


sschildhauer:
I am trying to add a class to the output of run_mod called "sr_mod_list" as recommended.  I then try to reference this class in the documentation of run_mod but it is throwing an error. I am wondering if there is any background work that needs to be done in order to establish this new class and reference it in other parts of the package? I searched online but am having a hard time finding any good answers. Thank you!

d-morrison:
> I am trying to add a class to the output of run_mod called "sr_mod_list" as recommended. I then try to reference this class in the documentation of run_mod but it is throwing an error. I am wondering if there is any background work that needs to be done in order to establish this new class and reference it in other parts of the package? I searched online but am having a hard time finding any good answers. Thank you!

@sschildhauer good question! we definitely want to document the class; I think the best way is to create a separate R file for the "sr_mod_list" class, similar to the one for `?tbl_df`: https://github.com/tidyverse/tibble/blob/main/R/tbl_df.R

you can skip lines 1-3 of that file for now. [lines 5-67](https://github.com/tidyverse/tibble/blob/fe2edfe2bcc3482e1451931dfe6a5bd4c6d9518e/R/tbl_df.R#L5-L67) are the key part; the contents will probably be a little different though, and we should include a hyperlinked list of all the functions you've written that take "sr_mod_list"s as inputs.

Glad to discuss details if helpful!

d-morrison (changes requested):
(I know this is still in-progress; just adding a review to get it out of my to-do list for now, and to make sure you're aware that there's another merge conflict to resolve)

d-morrison (approved):
Looks good - thanks!

PR #79, opened 2025-04-01
Updating run_mod output

sschildhauer:
Modifying the run_mod output to exclude raw jags.post data. The raw object made the output huge in size and may not be necessary for diagnostics, estimates, or other portions of the package. However, if anyone believes we should still keep the jags.post included in the output, I am happy to cancel this pull request. 

Curve_params will now include the posterior distribution of each modeled subject. This will hopefully make it easier to create further visualization and estimation functions that analyze the posterior of each subject, and not just the "new person" without having to unpack the mcmc object every time. 

I also modified the diagnostic and estimation functions to accommodate the new object. These functions now take the curve_params data set and filter down to only represent the "newperson".


sschildhauer:
Im a bit confused to why this is not passing some of the reviews. It says that the snap for plot_jags_effective in vidffr is failing. So I went and deleted the current snap and reran the test to create a new snap. However, it is still failing (but passing on macOS). Thank you for taking a look!

d-morrison:
> Im a bit confused to why this is not passing some of the reviews. It says that the snap for plot_jags_effective in vidffr is failing. So I went and deleted the current snap and reran the test to create a new snap. However, it is still failing (but passing on macOS). Thank you for taking a look!

@sschildhauer good question; operating system-dependent differences can be tricky, and it's especially weird that the check on ubuntu-latest (old-rel) is passing but ubuntu (release) and (devel) are failing.

You can extract the changed snapshots from the github artifact; see the link in the last line of this snapshot: 

image

You can inspect the files visually, or if you copy `typhoid-plot-ess.new.svg` into `serodynamics/tests/testthat/_snaps/plot_jags_effectivedx` on your computer, you can then run `testthat::snapshot_review()` as usual.

In this case, it looks like relative ordering of `shape` and `t1` is changing on the y-axis:

New:
typhoid-plot-ess new

Old:
typhoid-plot-ess

At first glance I'm not sure why that would happen; I've run into issues with inconsistent string ordering before, which I've solved by adding this line to the `.Rprofile` file: 
`Sys.setlocale(category = "LC_COLLATE", locale = "en_US.UTF-8")`

That line is already in the .Rprofile for serodynamics (you can inspect it with `usethis::edit_r_profile(scope = "project")`).

I'm stumped at the moment. Please double check the steps I did above to make sure I did those checks correctly, and then maybe try running the tests on Posit Cloud (which uses Ubuntu) to see if you get the same issue there. If you're still stuck after that, let's discuss at the meeting tomorrow evening.

@Kwan-Jenny please also take a look and see if you can spot the issue?

sschildhauer:
All reviews are now passing. 

> > Im a bit confused to why this is not passing some of the reviews. It says that the snap for plot_jags_effective in vidffr is failing. So I went and deleted the current snap and reran the test to create a new snap. However, it is still failing (but passing on macOS). Thank you for taking a look!
> 
> @sschildhauer good question; operating system-dependent differences can be tricky, and it's especially weird that the check on ubuntu-latest (old-rel) is passing but ubuntu (release) and (devel) are failing.
> 
> You can extract the changed snapshots from the github artifact; see the link in the last line of this snapshot:
> 
> image
> 
> You can inspect the files visually, or if you copy `typhoid-plot-ess.new.svg` into `serodynamics/tests/testthat/_snaps/plot_jags_effectivedx` on your computer, you can then run `testthat::snapshot_review()` as usual.
> 
> In this case, it looks like relative ordering of `shape` and `t1` is changing on the y-axis:
> 
> New: typhoid-plot-ess new
> 
> Old: typhoid-plot-ess
> 
> At first glance I'm not sure why that would happen; I've run into issues with inconsistent string ordering before, which I've solved by adding this line to the `.Rprofile` file: `Sys.setlocale(category = "LC_COLLATE", locale = "en_US.UTF-8")`
> 
> That line is already in the .Rprofile for serodynamics (you can inspect it with `usethis::edit_r_profile(scope = "project")`).
> 
> I'm stumped at the moment. Please double check the steps I did above to make sure I did those checks correctly, and then maybe try running the tests on Posit Cloud (which uses Ubuntu) to see if you get the same issue there. If you're still stuck after that, let's discuss at the meeting tomorrow evening.
> 
> @Kwan-Jenny please also take a look and see if you can spot the issue?

Thank you for the thorough reply. I went ahead and ordered the y-axis explicitly, so this should not happen again.

sschildhauer:
Passing all reviews. Thank you for taking a look!

data-raw/nepal_sees_jags_output.R
```diff
@@ -0,0 +1,18 @@
+dataset <- serodynamics::nepal_sees |>
+  as_case_data(id_var = "person_id",
+               biomarker_var = "antigen_iso",
+               value_var = "result",
+               time_in_days = "dayssincefeveronset")
```
d-morrison, line 5: let's incorporate this step into the definition of the `nepal_sees` object (in `data-raw/nepal_sees.R`)?

data-raw/nepal_sees_jags_output.R
d-morrison, line 1: if this file supersedes `data-raw/nepal_sees_jags_post.R`, let's delete that file, as well as the corresponding `nepal_sees_jags_post.rda` object in `data/`

inst/WORDLIST
```diff
@@ -30,6 +30,6 @@ serocourse
 seroresponse
 strat
 stratifications
-summ
+subects
```
d-morrison, line 33: are we sure we want to allow `subects`? seems like a typo to me

NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 ## New features
 
+* Taking jags.post out of run_mod output. Including all subects in
```
d-morrison, line 7:
```suggestion
* Taking jags.post out of run_mod output. Including all subjects in 
```

R/Run_Mod.R
```diff
@@ -161,12 +160,18 @@ run_mod <- function(data,
...
-    "curve_params" = jags_out,
-    "jags.post" = jags_post_final,
-    "attributes" = mod_atts
-  )
+  # jags output results for each stratification rbinded.
+  if (with_post == TRUE) {
+    jags_out <- list(
+      "curve_params" = jags_out,
+      "jags.post" = jags_post_final,
+      "attributes" = mod_atts
```
d-morrison, line 168: is one of these three elements the "main" output in some sense? for example, does one of them get extracted and further analyzed more than the others? if so, I'd recommend turning the other elements into metadata attributes (for example using `structure()`)

R/plot_jags_densitydx.R
```diff
@@ -58,7 +42,8 @@ plot_jags_dens <- function(data,
   for (i in strat) {
 
     visualize_jags_sub <- visualize_jags |>
-      dplyr::filter(.data$Stratification == i)
+      dplyr::filter(.data$Stratification == i) |>
+      dplyr::filter(.data$Subject == "newperson")
```
d-morrison, line 46: maybe if `visualize_jags` contains more than one value of `Subject`, we should loop over the subjects like we do for isos and strata?

d-morrison (changes requested):
@sschildhauer I know you're still working on this PR, but I had some time to look at it today and left some comments; glad to discuss!

R/Run_Mod.R
d-morrison, line 1: PS - just to make sure we don't lose track of it, I still think we should add an option to only export the samples for `newperson` (so that the previous behavior of this function is still available)

R/Run_Mod.R
d-morrison: (let's save this for a separate PR; see #83)

d-morrison (commented):
a few more comments

R/Run_Mod.R
sschildhauer: Okay, that sounds good. I can implement this in the next PR.

sschildhauer:
> > Modifying the run_mod output to exclude raw jags.post data. The raw object made the output huge in size and may not be necessary for diagnostics, estimates, or other portions of the package. However, if anyone believes we should still keep the jags.post included in the output, I am happy to cancel this pull request.
> > Curve_params will now include the posterior distribution of each modeled subject. This will hopefully make it easier to create further visualization and estimation functions that analyze the posterior of each subject, and not just the "new person" without having to unpack the mcmc object every time.
> > I also modified the diagnostic and estimation functions to accommodate the new object. These functions now take the curve_params data set and filter down to only represent the "newperson".
> 
> Taking another look at this description, I'm realizing that part of the reason this PR got so complicated is that it's addressing two separate issues: 1) adding the option to exclude the raw `jags.post` data from the output, and 2) including all subjects in the mcmc sample output; I'll accept this double-PR for now, but in the future I'll insist on keeping separate issues in separate PRs

I agree.. I think I should have compartmentalized this a bit better.

R/plot_jags_densitydx.R
sschildhauer: I agree. I think I will do this in another pull request.

sschildhauer:
I implemented many of the recommended changes. A couple of quick notes:
- I will make the diagnostic functions modifiable to specify any subject with a default of newperson in another pull request. 
- I added the option to include jags.post. I agree that curve_params will be the primary object people will work with. So, I will make the other objects meta data in another PR.
- I added an option to include all subjects or just the newperson. I probably should have done this in another PR but noticed after the fact.

d-morrison (changes requested):
(see responses in thread above)

d-morrison (changes requested):
Almost there! Please see inline comments.

R/plot_jags_densitydx.R
d-morrison: > I agree. I think I will do this in another pull request.

(tracked in issue https://github.com/UCD-SERG/serodynamics/issues/84#issuecomment-2803370315)

pkgdown/_pkgdown.yml
```diff
@@ -35,6 +35,7 @@ reference:
   - plot_jags_Rhat
   - plot_jags_trace
   - plot_jags_effect
+  - plot_title_fun
```
d-morrison, line 38:
```suggestion
```
I don't think we need to include `plot_title_fun` in the reference index; it's not an exported function

R/Run_Mod.R
```diff
@@ -21,11 +21,11 @@
 #' @param niter An [integer] specifying number of iterations.
 #' @param strat A [character] string specifying the stratification variable,
 #' entered in quotes.
-#' @param with_post A [logical] value specifying if raw jags post objects
+#' @param with_post A [logical] value specifying whether raw `jags.post` objects
 #' should be included in the output. Note: These objects can be large.
```
d-morrison, line 25:
```suggestion
#' @param with_post A [logical] value specifying whether a raw `jags.post` component
#' should be included as an element of the [list] object returned by `run_mod()` 
#' (see `Value` section below for details).
#' Note: These objects can be large.
```

R/nepal_sees_jags_output.R
```diff
@@ -6,14 +6,16 @@
 #' using the [nepal_sees] example data set as input
 #' and stratifying by column `"bldculres"`, 
 #' which is the diagnosis type (typhoid or
-#' paratyphoid). Keeping only the newperson, sees_npl_1, sees_npl_2.
+#' paratyphoid). Keeping only IDs "newperson", "sees_npl_1", "sees_npl_2".
```
d-morrison, line 9:
```suggestion
#' paratyphoid). Keeping only IDs `"newperson"`, `"sees_npl_1"`, and `"sees_npl_2"`.
```

R/nepal_sees_jags_output.R
```diff
@@ -6,14 +6,16 @@
...
 #' @format ## `nepal_sees_jags_output`
 #' A [list] consisting of the following named elements:
 #' \describe{
 #'  \item{curve_params}{A [data.frame] titled `curve_params` that contains the
-#'   posterior distribution for the predictive distribution (newperson) and 
-#'   two modeled subject (sees_npl_1, sees_npl_2)}
+#'   posterior predictive distribution of the person-specific parameters for a
+#'   "new person" with no observed data (`Subject = "newperson"`) and posterior
+#'   distributions of the person-specific parameters for two arbitrarily-chosen
+#'   subjects (`"sees_npl_1"` and`"sees_npl_2"`)}
```
d-morrison, line 18:
```suggestion
#'   subjects (`"sees_npl_1"` and `"sees_npl_2"`)}
```

d-morrison (changes requested):
A few more notes - almost there!

sschildhauer:
Im not totally sure why this one isn't passing the last reviews. Thank you for taking another look! Apologies for all of the issues.

R/Run_Mod.R
```diff
@@ -21,12 +21,17 @@
...
+#' should be included as an element of the [list] object returned by `run_mod()`
+#' (see `Value` section below for details).
+#' Note: These objects can be large.
 #' @param include_subs A [logical] value specifying whether posterior
 #' distributions should be included for all subjects. A value of [FALSE] will
 #' only include the predictive distribution.
-#' @return
+#' @returns A list containing the following elements:
+#' - A jags.post [list()] object or multiple jags.post [list()]
+#' if stratified. Returned as a [list()] of class [runjags::runjags-class]
```
d-morrison, line 34: see https://github.com/UCD-SERG/serodynamics/pull/79#discussion_r2049779605; this information should be more precise, and the sentence structure seems a bit off

R/Run_Mod.R
```diff
@@ -21,12 +21,17 @@
...
+#' @param with_post A [logical] value specifying whether a raw `jags.post`
+#' component
+#' should be included as an element of the [list] object returned by `run_mod()`
+#' (see `Value` section below for details).
+#' Note: These objects can be large.
 #' @param include_subs A [logical] value specifying whether posterior
 #' distributions should be included for all subjects. A value of [FALSE] will
 #' only include the predictive distribution.
-#' @return
+#' @returns A list containing the following elements:
```
d-morrison, line 32:
```suggestion
#' @returns A [list] containing the following elements:
```

d-morrison: it looks like these brackets got lost from my [previous review](https://github.com/UCD-SERG/serodynamics/pull/79#discussion_r2049780305) (unless you intended to omit them); 
going forward, please use the "commit suggestion" button when possible to avoid this issue:
image

d-morrison (changes requested):
almost there, please see comments

R/Run_Mod.R
sschildhauer: Ahh okay, I will make sure to commit suggestions from here on out.

R/Run_Mod.R
```diff
@@ -29,9 +29,8 @@
 #' @param include_subs A [logical] value specifying whether posterior
 #' distributions should be included for all subjects. A value of [FALSE] will
 #' only include the predictive distribution.
-#' @returns A list containing the following elements:
-#' - A jags.post [list()] object or multiple jags.post [list()]
-#' if stratified. Returned as a [list()] of class [runjags::runjags-class]
+#' @returns A [list] containing the following elements:
+#' - A jags.post [list] returned as class [runjags::runjags-class].
```
d-morrison, line 33:
```suggestion
#' - `"jags.post"`: a [list] containing one or more [runjags::runjags-class] objects (one per stratum)
```

R/Run_Mod.R
```diff
@@ -19,12 +19,18 @@
...
+#' (see `Value` section below for details).
+#' Note: These objects can be large.
+#' @param include_subs A [logical] value specifying whether posterior
+#' distributions should be included for all subjects. A value of [FALSE] will
+#' only include the predictive distribution.
+#' @returns A [list] containing the following elements:
+#' - A jags.post [list] returned as class [runjags::runjags-class].
 #' - A [base::data.frame()] titled `curve_params` that contains the posterior
 #' distribution will be exported with the following attributes:
 #'  - `iteration` = number of sampling iterations.
```
d-morrison, line 36: check the [preview documentation](https://ucd-serg.github.io/serodynamics/preview/pr79/reference/run_mod.html); these second-level bullet points need more indentation to display correctly:

image

d-morrison (changes requested):
almost there; please look through all the changed documentation to make sure everything looks as expected (see comment)

R/Run_Mod.R
sschildhauer: Got it. Added second level bullet points.

sschildhauer:
Hi Ezra, I went through the documentation and fine tuned everything. Thank you for emphasizing this practice -- I need to edit the documentation output rather than the code to pick up on small errors.

R/Run_Mod.R
```diff
@@ -13,12 +13,12 @@
 #'  - alpha = decay rate
 #' @param data A [base::data.frame()] with the following columns.
 #' @param file_mod The name of the file that contains model structure.
-#' @param nchain An [integer] between 1 and 4 that specifies
+#' @param nchain An [integer] between 1 and 4 that specifies.
 #' the number of mcmc chains to be run per jags model.
```
d-morrison, line 17:
```suggestion
#' @param nchain An [integer] between 1 and 4 that specifies
#' the number of mcmc chains to be run per jags model.
```

d-morrison:
> Hi Ezra, I went through the documentation and fine tuned everything. Thank you for emphasizing this practice -- I need to edit the documentation output rather than the code to pick up on small errors.

I'm glad you're seeing the method behind the madness; I know it's a lot of small detail, but it's a lot easier to take care of it now rather than letting ["technical debt"](https://en.wikipedia.org/wiki/Technical_debt) accumulate over time. 

There are actually still some issues in the `run_mod()` documentation that we should eventually address, but they aren't in the lines that you've edited so far, so we can skip them for now. I recorded them in issue #96.

d-morrison (changes requested):
I saw just one more typo in the lines you've been editing; just a stray period. Other than that, I think we can merge this PR and move on.

d-morrison (approved):
Looks good - thanks for all the hard work!

PR #99, opened 2025-04-27
Changes to run_mod output.

sschildhauer:
This pull request is designed to clean up the run_mod output:
- Making the resulting object a tbl. It is also the only output now.
- Making all other pieces attributes including the model parameter inputs and the optional jags.post. 
- Taking out include_subs as an option -- instead including all individuals as a default.

Apologies for putting this all into one pull request -- it felt appropriate given that they are all modifications to the output and it might have been tough to do some without others.


sschildhauer:
I also fixed capitalization in documentation corresponding to issue an issue that was opened.

R/Run_Mod.R
```diff
@@ -127,13 +125,13 @@ run_mod <- function(data,
     # stratification and will only be included if specified. 
     jags_post_final[[i]] <- jags_post
 
-    # Unpacking and cleaning mcmc output.
+    # Unpacking and cleaning MCMC output.
     jags_unpack <- ggmcmc::ggs(jags_post[["mcmc"]])
 
     # Adding attributes
     mod_atts <- attributes(jags_unpack)
     # Only keeping necesarry attributes
```
Kwan-Jenny, line 133:
```suggestion
    # Only keeping necessary attributes
```

R/Run_Mod.R
```diff
@@ -26,31 +26,30 @@
...
-#'   - `y1` = posterior estimate of peak antibody concentration
-#'   - `stratified variable` = the variable used to stratify jags model
-#' - A [list] of `attributes` that summarize the jags inputs, including:
+#' @returns A [dplyr::tbl_df] that contains the posterior
+#' distribution will be exported with the following variables:
+#'   - `iteration` = Number of sampling iterations
+#'   - `chain` = Number of MCMC chains run; between 1 and 4
+#'   - `Parameter` = Parameter being estimated. Includes the following:
+#'     - `alpha` = Posterior estimate of decay rate
+#'     - `r` = Posterior estimate of shape parameter
```
Kwan-Jenny, line 35: This is just to confirm that the 'r' parameter corresponds to 'shape' in the serocalculator package, if I remember correctly.
Do we need to use the same parameter names in both packages?

R/Run_Mod.R
```diff
@@ -169,25 +167,21 @@ run_mod <- function(data,
...
-    )
-  } else { 
-    jags_out <- list(
-      "curve_params" = jags_out,
-      "attributes" = mod_atts
-    )
-  }
+    jags_out <- jags_out |>
+      structure(jags.post = jags_post_final)
+  }
```
Kwan-Jenny, line 185: I noticed that when adding the jags.post attribute, the code uses structure(jags.post = jags_post_final).
I’m wondering if it might be safer to preserve the existing attributes (like nChain, nIterations, etc.) when adding jags.post.
One possible way could be something like:
```suggestion
jags_out <- structure(.Data = jags_out, 
                      .Attributes = c(attributes(jags_out), list(jags.post = jags_post_final)))
```
I’m not 100% sure if it’s necessary here, but it might help avoid accidentally losing important metadata. Just an idea — feel free to ignore if you think the current approach already works fine!

R/Run_Mod.R
sschildhauer: Agreed! Good catch. Changing "r" to shape throughout documentation. It is output as "shape" as well.

R/Run_Mod.R
sschildhauer: It looks like this adds on the attributes again with an additional $Attributes required to call them. I was trying to structure the attributes so that the output could be called by "attributes(output_object)$attribute_of_interest", while this would require "attributes(output_object)$Attributes$attribute_of_interest". For now, I think we should keep as it and consult with Ezra.

R/Run_Mod.R
```diff
@@ -26,31 +26,30 @@
...
+#'     - `y0` = Posterior estimate of baseline antibody concentration
+#'     - `y1` = Posterior estimate of peak antibody concentration
+#'     - `t1` = Posterior estimate of time to peak
+#'     - `shape` = Posterior estimate of shape parameter
+#'     - `alpha` = Posterior estimate of decay rate
+#'   - `Iso_type` = Antibody/antigen type combination being evaluated
+#'   - `Stratification` = The variable used to stratify jags model
+#'   - `Subject` = ID of subject being evaluated
+#'   - `value` = Estimated value of the parameter
+#' - The following `attributes` are included in the output:
```
d-morrison, line 43:
```suggestion
#' - The following [attributes] are included in the output:
```

R/Run_Mod.R
```diff
@@ -26,31 +26,30 @@
...
 #'   - `class`: Class of the output object.
 #'   - `nChain`: Number of chains run.
 #'   - `nParameters`: The amount of parameters estimated in the model.
 #'   - `nIterations`: Number of iteration specified.
 #'   - `nBurnin`: Number of burn ins.
 #'   - `nThin`: Thinning number (niter/nmc).
+#'   - An optional `"jags.post"` included when argument `with_post` = TRUE.
```
d-morrison, line 50:
```suggestion
#'   - An optional `"jags.post"` attribute, included when argument `with_post` = TRUE.
```

d-morrison (changes requested):
marked a couple of documentation edits; otherwise, LGTM!

sschildhauer:
Thank you for the additional comments! All suggestions committed.

d-morrison (approved):
Looks good!

PR #114, opened 2025-07-09
Creating a default print for sr_model

sschildhauer:
Three CI failures in the `print.sr_model` PR: tidyselect deprecation warnings from `.data` in non-data-masking contexts, Rd files out of sync with source, and a missing snapshot for the unstratified test case.

## Changes

- **Fix `.data` in tidyselect contexts** — `.by`, `names_from`, and `values_from` are tidyselect args, not data-masking; replaced with string column names:
  ```r
  # Before
  dplyr::summarise(.by = c(.data$Stratification, .data$Iso_type, .data$Parameter), ...)
  tidyr::pivot_wider(names_from = .data$Parameter, values_from = .data$median_val)

  # After
  dplyr::summarise(.by = c("Stratification", "Iso_type", "Parameter"), ...)
  tidyr::pivot_wider(names_from = "Parameter", values_from = "median_val")
  ```

- **Regenerate documentation** — ran `devtools::document()` to sync `man/print.sr_model.Rd` and four other Rd files that had drifted from source; also fixed `[dplyr::tbl_df]` → `[tibble::tbl_df]` in the `print_tbl` param doc.

- **Add missing snapshot** — added the Linux/non-Darwin snapshot for the unstratified `sr_model` print test (the macOS variant will be generated by CI on the first passing run).


d-morrison:
> I am a little confused with the warning/error I am getting here. It looks like there is a scoping issue relating to the base dataset when using dplyr, but when I fix the issue the tests are producing a warning. It seems like the solution might be in conflicting with each other. Grateful for any help!

@sschildhauer I think this is the key error message: 

> Use of .data in tidyselect expressions was deprecated in tidyselect 1.2.0.
> i Please use `"mean_val"` instead of `.data$mean_val`

See https://dplyr.tidyverse.org/articles/programming.html for details?

NEWS.md
```diff
@@ -4,6 +4,7 @@
 
 ## New features
 
+* Creating default print.sr_model function for sr_model class (#)
```
d-morrison, line 7:
```suggestion
* Creating default print.sr_model function for sr_model class (#114)
```

R/print.sr_model.R
```diff
@@ -0,0 +1,27 @@
...
+#'  `Stratification` (if specified).
+#' @param data An `sr_model` output object from the
+#' [serodynamics::run_mod()] function.
+#' @returns A [dplyr::grouped_df] that
+#' contains the mean posterior distribution for antibody kinetic curve
+#' parameters by `Iso_type` and `Stratification` (if specified).
+#' @export
+#' @examples
+#' print(nepal_sees_jags_output)
+print.sr_model <- function(data) { # nolint
```
d-morrison, line 15: I think you'll need to include the standard parameters for `print()` methods, `x` and `...`,
but you can wait until the automated tools tell you it's a problem.

See https://adv-r.hadley.nz/s3.html#s3-methods for details.

R/print.sr_model.R
```diff
@@ -0,0 +1,27 @@
...
+  data_group <- data |>
+    dplyr::group_by(.data$Stratification, .data$Iso_type, .data$Parameter) |>
+    dplyr::summarise(mean_val = mean(.data$value)) |>
+    tidyr::pivot_wider(names_from = .data$Parameter, 
+                       values_from = .data$mean_val) |>
+    dplyr::arrange(.data$Iso_type)
+  # Taking out stratification column if not specified
+  if (unique(data$Stratification == "None")) {
+    data_group <- data_group[, 2:7]
```
d-morrison, line 24: using numeric indices to select columns makes code hard for others (or your future self) to understand,
because they have to determine which columns those indices refer to.

Why not use `data_group <- data_group |> select(-all_of("Stratification"))` or similar?

R/print.sr_model.R
```diff
@@ -0,0 +1,27 @@
...
+  } 
+  data_group
```
d-morrison, line 26: print methods generally return the original input object, invisibly; check out `print.data.frame` for example.

tests/testthat/test-print.sr_model.R
```diff
@@ -0,0 +1,9 @@
+test_that(
+  desc = "results are consistent with printed output for sr_model class",
+  code = {
+    withr::local_seed(1)
+    results <- print(nepal_sees_jags_output) |>
+      ssdtools:::expect_snapshot_data("default-print-runmod-object")
```
d-morrison, line 6: for `print()` methods, I think we should use `expect_snapshot()`, which captures what actually gets printed in the console.

d-morrison (changes requested):
Please see comment above about tidyselect vs data-masking, and inline comments below.

R/print.sr_model.R
sschildhauer: Changed the input paramaters as recommended. Will keep an eye out for further automated tool checks.

R/print.sr_model.R
sschildhauer: I am having a hard time referencing some of the other print functions because it seems like our is a little different. We will see how my change works.

R/print.sr_model.R
```diff
@@ -0,0 +1,28 @@
...
+#' @returns A [dplyr::grouped_df] that
+#' contains the mean posterior distribution for antibody kinetic curve
+#' parameters by `Iso_type` and `Stratification` (if specified).
+#' @export
+#' @examples
+#' print(nepal_sees_jags_output)
+print.sr_model <- function(x, ...) { # nolint
+  x <- x |>
+    dplyr::group_by(.data$Stratification, .data$Iso_type, .data$Parameter) |>
+    dplyr::summarise(mean_val = mean(.data$value)) |>
```
d-morrison, line 19: @sschildhauer I'm wondering, are we actually trying to build a `summary()` method here instead of a `print()` method? maybe that's why this feels so different from other `print()` methods

sschildhauer: @d-morrison  @kaiemjoy That is a good question and bringing in Kristen on this. We were trying to come up with a default print that isn't just the full output tbl, and thought it could follow somewhat like other functions (like glm), where just the point estimate is shown when for the default print. I could make the default print have a bit more information -- like priors used, or bayesian inputs like number of samples, chains, etc. I was thinking the summary might be a bit more in depth, showing 95% credible intervals, maybe the mean and median of each value. However, I am open to what you think might be best.

sschildhauer:
Hi Ezra, 

I am having a harder time with this PR. I am not totally sure why one of the tests is not passing. Can you take a look when you have time? Also would be happy to hop on a quick chat to discuss.

Thank you,
Sam

sschildhauer:
> Hi Ezra,
> 
> I am having a harder time with this PR. I am not totally sure why one of the tests is not passing. Can you take a look when you have time? Also would be happy to hop on a quick chat to discuss.
> 
> Thank you, Sam

It looks like the test snap is recording different values than what are produced when going through github reviews. I am not totally sure why this would be happening -- maybe the median() command is being processed differently in the github reviews. @d-morrison Any pointers are greatly appreciated here!

d-morrison (changes requested):
@Kwan-Jenny please take a look and see if you can help Sam debug? if you're both stuck you can send it back to me

sschildhauer:
Hi Ezra and Kwan. I checked the snaps and having a hard time finding a solution. It looks like the test for print.sr_model in the github review is producing different median values than when I run the function locally. There also seems to be something going on with autoplot.case_data documention -- this is a new issue and has me stumped as well. Thank you so much for review. These should almost be there!

d-morrison (changes requested):
@sschildhauer re: the `document` error: `ggplot2` recently updated to v4.0 and changed a lot of stuff, including documentation, which the documentation for `autoplot.case_data()` depends on.

Please update `ggplot2` and its dependencies on your local machine and re-run `document()`; once you do that, the changes you introduced in `autoplot.case_data.Rd` will hopefully go away.

d-morrison:
@sschildhauer PS - I bet updating ggplot2 will also resolve the snapshot discrepancies

R/print.sr_model.R
```diff
@@ -0,0 +1,42 @@
...
+                               .data$Parameter), 
+                       median_val = stats::median(.data$value)) |>
+      tidyr::pivot_wider(names_from = .data$Parameter, 
+                         values_from = .data$median_val) |>
+      dplyr::arrange(.data$Iso_type) |>
+      suppressWarnings()
+    # Taking out stratification column if not specified
+    if (unique(x$Stratification == "None")) {
+      x <- x |> select(-c(.data$Stratification))
+    }
```
Kwan-Jenny: `unique(x$Stratification == "None")` can return a vector and cause unexpected results.
Maybe replace with something safer like:
```suggestion
    if (!"Stratification" %in% names(x) || all(x$Stratification == "None", na.rm = TRUE)) {
  x <- dplyr::select(x, -dplyr::any_of("Stratification"))
}
```

R/print.sr_model.R
```diff
@@ -0,0 +1,42 @@
...
+#' @description
+#'  A default print method for class `sr_model` that includes the median
+#'  posterior distribution for antibody kinetic curve parameters by `Iso_type`
+#'  and `Stratification` (if specified).
+#' @param x An `sr_model` output object from [run_mod()].
+#' @param print_tbl A [logical] indicator to print in style of [dplyr::tbl_df].
+#' @param ... Additional arguments affecting the summary produced.
+#' [serodynamics::run_mod()] function.
+#' @returns A data summary that
+#' contains the mean posterior distribution for antibody kinetic curve
```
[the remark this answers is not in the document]
d-morrison: @Kwan-Jenny actually the r-pkgs book recommends `@returns`: https://r-pkgs.org/man.html#fn10:~:text=return%2C%20but-,we%20now%20favor,-%40returns%20because%20it

sschildhauer:
@copilot please provide suggestions regarding how to fix the documentation error that I am receiving in the review. It looks like some Rd. files have been changed but when I try to run devtools::document() in the R interface there are no recommended updates.

Kwan-Jenny (commented):
Overall looks good and the tidyselect fix is clean! Just a couple of small things I noticed:

- In `print.sr_model`, the whole pipe is wrapped in `suppressWarnings()` — would be helpful to have a quick comment on what warning is being suppressed, just so it's easier to debug later if something breaks silently.

- Is `"None"` a sentinel value coming from the model output? If so, maybe worth a short comment explaining where it comes from so it doesn't look like a magic string.

- In the test file, `suppressWarnings()` is used inside the third test too — same thing, might be worth being a bit more explicit about what's expected there.

Minor stuff overall, nothing blocking.

d-morrison:
@copilot resolve the merge conflicts in this pull request

d-morrison (changes requested):
@sschildhauer please review the comments in [this thread](https://github.com/ucdavis/serodynamics/pull/114#pullrequestreview-4241653966)

keep iterating with copilot until it doesn't have any new suggestions

sschildhauer:
@copilot apply changes based on the comments in [this thread](https://github.com/ucdavis/serodynamics/pull/114#pullrequestreview-4241508406)

sschildhauer:
@copilot apply changes based on the comments in [this thread](https://github.com/ucdavis/serodynamics/pull/114#pullrequestreview-4248463687)

R/print.sr_model.R
```diff
@@ -0,0 +1,56 @@
...
+#' @return Invisibly returns either:
+#' \itemize{
+#'   \item When `print_tbl = TRUE`, a tibble containing the raw `sr_model`
+#'   draws;
+#'   \item Otherwise, a data summary containing posterior medians for
+#'   antibody kinetic curve parameters by `Iso_type` and `Stratification`
+#'   (if specified).
+#' }
+#' @export
+#' @examples
```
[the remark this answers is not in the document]
sschildhauer: @copilot apply changes based on [this feedback](https://github.com/ucdavis/serodynamics/pull/114#discussion_r3209352796)

sschildhauer:
@copilot apply changes based on the comments in [this thread](https://github.com/ucdavis/serodynamics/pull/114#pullrequestreview-4254428983)

PR #116, opened 2025-07-14
vectorize `ab()`

d-morrison:
@Kwan-Jenny this change should enable you to remove the the `rowwise()` and `ungroup()` calls when you use `ab()` in `plot_predicted_curve()`


Kwan-Jenny (approved):
The new version replaces the `if` statement with `ifelse()`, which makes the function vectorized. This means it can handle multiple `t` values at once instead of using loops, which is much faster for large datasets. It also removes the extra variable initialization and combines logic into one line, making the code cleaner and easier to maintain. The main reason for this change is efficiency, better readability, and flexibility for simulations or plotting curves.

PR #121, opened 2025-07-28
Adding ability to run diagnostic functions on individuals

sschildhauer:
This pull request will add the ability to run diagnostic functions on individuals using "id" as an input. Apologies for all of the files changed and this relatively big PR -- the same small changes were made across many functions, so I felt it made sense to group it into one PR. A couple of details:
- The default is "newperson".
- You can produce results for multiple individuals at once.
- The results produce lists of lists. I know this is not ideal, but I felt that facetwrap would likely create too convoluted figures. Plus, the list design makes it easy to call a specific figure from the result. We can discuss alternatives.

### DIFF
--- a/tests/testthat/test-run_mod.R
+++ b/tests/testthat/test-run_mod.R
@@ -149,7 +149,7 @@ test_that(
 
 test_that(
   desc = "results are consistent with unstratified SEES data with modified 
-  priors and additional parameters",
+  priors",
   code = {
     announce_snapshot_file("nostrat-curve-params-specpriors.csv")
     skip_on_os(c("windows", "linux"))


d-morrison:
@sschildhauer please resolve conflicts
@Kwan-Jenny please review line-by-line and look for opportunities to simplify code and/or API
@kaiemjoy please beta-test and assess whether list-of-lists output works for you?

R/plot_jags_tracedx.R
```diff
@@ -34,50 +37,59 @@
...
-      dplyr::filter(.data$Stratification == i) |>
-      dplyr::filter(.data$Subject == "newperson")
+      dplyr::filter(.data$Subject == h)
+    
+    trace_strat_list <- list()
+  
+    for (i in strat) {
+
+      visualize_jags_sub <- data |>
+        dplyr::filter(.data$Stratification == i)
```
Kwan-Jenny, line 56: It looks like we start from the full `data` again inside the strat loop, so the earlier `Subject == h` filter can be lost. If we continue filtering from the already-subset data, the id argument will always be respected.

```suggestion
visualize_jags_sub <- visualize_jags_sub |>
  dplyr::filter(.data$Stratification == i)
```

PR #216, opened 2026-05-15
Harden Claude code-review workflow against races and silent failures

d-morrison:
## Summary

Backports review feedback from [d-morrison/rme#739](https://github.com/d-morrison/rme/pull/739), where Claude reviewed a parallel port of this same workflow. The same issues exist here.

- **Add a `concurrency:` group** so two overlapping PR runs can't race the stash/restore and silently strip reviewers from the PR. `cancel-in-progress: true` so the freshest diff wins.
- **Drop `|| true` on the restore POST** — a failed restore means the human reviewer was silently dropped and never gets the post-Claude notification, defeating the serialization signal. Fail loudly so it can be re-added manually instead of vanishing.
- **Document the intent of the stash/restore inline** — it's a serialization signal (clear reviewers so they don't get notified mid-run; restore so the re-add fires a fresh notification once Claude is done), not a noop.
- **Pass `${{ github.repository }}` via `REPO` env** rather than inlining the GitHub expression into shell scripts in three places.
- **Replace `head -n 1` with a `while read` loop** when deleting prior Claude top-level comments — PRs predating `use_sticky_comment` may have multiple stale comments, and the prior code only cleaned one per run.
- **Remove dead `USERS=${USERS:-[]}` defaults** — the stash step always writes valid JSON to `GITHUB_OUTPUT`, so the fallback was unreachable.

## Test plan

- [ ] Push two commits in quick succession to a draft PR; confirm the second run cancels the first (`concurrency`) and reviewers end up correctly restored at the end.
- [ ] On a PR with a human reviewer requested, confirm the reviewer is removed during the run and re-added afterward (notification fires).
- [ ] Force a restore failure (e.g. revoke the token mid-run) and confirm the workflow surfaces the failure rather than silently swallowing it.
- [ ] On a PR with multiple pre-existing Claude comments, confirm all are cleaned up in one run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -10,6 +10,15 @@ on:
     #   - "src/**/*.js"
     #   - "src/**/*.jsx"
 
+# Serialize runs per PR. The stash/restore reviewer dance below assumes
+# only one run is mutating the PR's reviewer list at a time; without
+# this, an overlapping second run reads an already-cleared list, stashes
+# [], and on restore wipes the original reviewers permanently. Cancel
+# the in-flight run on a new push so the freshest diff wins.
+concurrency:
+  group: claude-review-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
   claude-review:
     # Optional: Filter by PR author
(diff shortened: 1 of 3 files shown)


d-morrison:
@claude review

d-morrison:
@claude review

Issue #231, opened 2026-05-22
Cut down on run_mod tests and assign order


sschildhauer:
Minor: Tests 2 and 3 lack the strict attribute ordering assertion
tests/testthat/test-run_mod.R:102-107 (SEES data) and 150-156 (unstratified with pop params)

Test 1 now checks both expect_setequal(names(attributes(...))) AND the strict ordering:

expect_equal(names(attributes(results))[1:3], c("names", "row.names", "class"))
Tests 2 and 3 only use expect_setequal and do not verify ordering. Given that attribute ordering was the root cause of the original CI failures across many review rounds, and the ordering is explicitly enforced in Run_Mod.R:271-281, a one-line guard in tests 2 and 3 would be cheap insurance. Not a blocker — test 1 already proves the code path.

sschildhauer:
Issue closed because it was completed in PR #253

PR #253, opened 2026-06-22
Cutting down on and cleaning R tests

sschildhauer:
This PR is cutting down on unnecessary tests and consolidating tests to minimize run time. Currently tests run.jags 5 times, current changes make it so run.jags occur only 3 times:
- One test is created for simulated, stratified data that includes population parameters
- The second test uses stratified SEES data and includes a mcmc post object, population parameters, and tests for prec_log_y. 
- The third test uses unstratified SEES data with modifiable priors.


tests/testthat/test-run_serodynamics.R
```diff
@@ -27,11 +31,14 @@ test_that(
     ) |>
       suppressWarnings()
     
-    results |>
-      expect_snapshot_data(
-        "sim-strat-curve-params",
-        variant = darwin_variant()
-      )
+    if (system_os() == "darwin") {
```
d-morrison: why only on darwin?

imelainew (commented):
I'm still getting familiar with the test suite, so I want to make sure I understand the consolidation correctly: the three remaining tests cover (1) simulated stratified data, (2) stratified SEES data, and (3) unstratified SEES data. Were the tests you removed all already covered by these three? Just want to understand which scenario each test maps to.

d-morrison:
Working on the failing `test-coverage` job here — paws off until I'm done.

**Root cause:** commit `22996a7` ("Taking away darwin wrapper from first test") removed the `if (system_os() == "darwin")` guard around the `sim-strat-curve-params` snapshot, but no Linux/Windows-default snapshot is committed (only the `_snaps/darwin/...` variant). On the Ubuntu coverage runner `darwin_variant()` returns `NULL`, so testthat generates a brand-new snapshot and fails — which halts `covr::package_coverage()` before `cobertura.xml` is written, so the "No coverage reports found" Codecov error is just the downstream symptom.

**Fix (per @d-morrison):** drop the `if (system_os() == "darwin")` guards on all `run_serodynamics` snapshot assertions and rely on `variant = darwin_variant()` alone — the same pattern `test-runjags.R` already uses (both `_snaps/runjags/` and `_snaps/darwin/runjags/` committed). This actually exercises the snapshots on the Linux coverage job instead of skipping them. I'll commit the generated default-variant snapshots from the coverage run so CI goes green.

---

d-morrison:
@sschildhauer you already did the right **first half** — `31104ba` "Taking out all darwin wrappers" matches the intended fix exactly (drop the guards, keep `variant = darwin_variant()`, same pattern as `test-runjags.R`).

But it's only half, which is why the coverage job is still red ([run 1788](https://github.com/UCD-SERG/serodynamics/actions/runs/28485660570) failed the same way). With the guards gone, all six `run_serodynamics` snapshots now run on the Ubuntu coverage runner, but only the `_snaps/darwin/run_serodynamics/` variant is committed — there's no default (Linux/Windows) variant, so testthat generates six brand-new snapshots and fails with `Adding new file snapshot`. (That failure halts `covr::package_coverage()` before `cobertura.xml` is written, so the Codecov "No coverage reports found" error is just the downstream symptom.)

**Remaining step: commit the six Linux/Windows default snapshots.** You can't make these on your Mac — a local `snapshot_accept()` there regenerates the `darwin` variant you already have. The Linux runner already generated them, so pull them from the failing run's artifact:

1. Open the failed **test-coverage** run for `31104ba` → **Artifacts** → download **`coverage-test-failures`**.
2. Unzip and copy the `_snaps/run_serodynamics/` folder it contains (under `serodynamics/serodynamics-tests/testthat/`) into `tests/testthat/_snaps/run_serodynamics/` — six files: `sim-strat-curve-params.csv`, `sim-strat-fitted_residuals.csv`, `strat-curve-params.csv`, `popparam-strat-summary-stats.csv`, `strat-fitted_residuals.csv`, `nostrat-curve-params-specpriors.csv`.
3. Commit + push. The coverage job then compares against them and passes — same bootstrap `_snaps/runjags/` already uses.

I couldn't do this for you from my sandbox: the session can only push to its own branch (403 to `cleaning_tests`) and the artifact host is blocked by the egress policy, so I can't fetch the CSVs to hand over either. Done here — paws back to you.

---

sschildhauer:
> I'm still getting familiar with the test suite, so I want to make sure I understand the consolidation correctly: the three remaining tests cover (1) simulated stratified data, (2) stratified SEES data, and (3) unstratified SEES data. Were the tests you removed all already covered by these three? Just want to understand which scenario each test maps to.

Hi @imelainew, I consolidated the tests to make sure all binary options were tested positive in at least one of the tests. We had some repetitive tests before and options somewhat scattered across different tests (ex. pop_params = T, with_post = F), so, this update should consolidate these.

d-morrison (approved):
LGTM!

sschildhauer:
Related to issue: #231

PR #255, opened 2026-06-26
fix: force LF on shell scripts so shebangs don't break (CRLF)

### LINKED ISSUE #254: Force LF line endings on shell scripts (.gitattributes)
Shell scripts checked out with Windows CRLF line endings get a `#!/bin/bash\r` shebang, so the kernel looks for the interpreter `/bin/bash\r` (shown as `/bin/bash^M`) and fails with `bad interpreter: No such file or directory`. `core.autocrlf` only normalizes on commit, not on checkout, so a working tree that picks up CRLF stays broken.

Add a `.gitattributes` rule `*.sh text eol=lf` so git forces LF in the working tree on every checkout, regardless of each user's global git config.

(Same fix applied in the sparta repo.)

d-morrison:
Closes #254.

A shell script checked out with Windows CRLF line endings gets a `#!/bin/bash\r` shebang; the kernel then looks for the interpreter `/bin/bash\r` (shown as `/bin/bash^M`) and fails with `bad interpreter: No such file or directory`. `core.autocrlf` normalizes CRLF only on **commit**, never on **checkout**, so a working tree that picks up CRLF stays broken.

This adds `.gitattributes` with `*.sh text eol=lf`, which forces LF in the working tree on every checkout regardless of each user's global git config. No committed shell scripts in this repo currently have CRLF, so this is purely preventive — no file renormalization.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,7 @@
+# Auto-detect text files and normalize them to LF in the working tree on
+# checkout, regardless of each contributor's global git config.
+* text=auto eol=lf
+
 # Shell scripts must use LF line endings — a CRLF on the shebang line makes
 # the kernel look for "/bin/bash\r" and fail with "bad interpreter".
 *.sh text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# Shell scripts must use LF line endings — a CRLF on the shebang line makes
+# the kernel look for "/bin/bash\r" and fail with "bad interpreter".
+*.sh text eol=lf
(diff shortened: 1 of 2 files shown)


d-morrison:
**ARDI status.** Changelog now passes (the `no changelog` label is applied; it took a fresh push to re-evaluate — re-running the old run replayed the pre-label payload). version-check passes after the dev-version bump.

The one remaining red is **windows-latest (release)**, a required check, failing at `setup-r@v2` with `Failed to get R 4.6.0: ... 404`. This is an infrastructure issue — R 4.6.0's Windows binary/manifest isn't resolving — and it hits every PR right now (R-CMD-check on `main` passed 3 days ago). It's unrelated to this `.gitattributes`/version change. It should clear once the R 4.6.0 Windows toolchain is available; otherwise this needs an admin-merge.

Issue #266, opened 2026-07-06
Temporarily disable automatic claude-review workflow


d-morrison:
Disable the automatic `pull_request`-triggered Claude review (`.github/workflows/claude-code-review.yml`) for now, without removing the workflow file, so it can be re-enabled later.

Keep the `workflow_dispatch` path intact so the review can still be manually triggered (e.g. by `claude.yml`'s re-dispatch after an `@claude` run pushes commits) if needed.

PR #284, opened 2026-08-11
No default priors

sschildhauer:
Taking away default priors and forcing users to input their priors manually. Changing documentation accordingly. Priors are now listed as NULL in the input for prep_priors().

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Internal
 
+* Took away default priors from `run_serodynamics()`. Users must manually 
+  specify priors now in order to run function.
 * Disabled the `@claude` agent bot.
   `.github/workflows/claude.yml`'s reactive triggers are commented out and its
   job carries `if: false`, so no comment, issue, or review event invokes the
--- a/R/prep_priors.R
+++ b/R/prep_priors.R
@@ -8,27 +8,24 @@
 #' @param mu_hyp_param A [numeric] [vector] of 5 values representing the prior
 #' mean for the population level parameters
 #' parameters (y0, y1, t1, r, alpha) for each biomarker.
-#' If specified, must be 5 values long, representing the following parameters:
-#'    - y0 = baseline antibody concentration (default = 1.0)
-#'    - y1 = peak antibody concentration (default = 7.0)
-#'    - t1 = time to peak (default = 1.0)
-#'    - r = shape parameter (default = -4.0)
-#'    - alpha = decay rate (default = -1.0)
+#' Must be 5 values long, representing the following parameters:
+#'    - y0 = baseline antibody concentration
+#'    - y1 = peak antibody concentration
+#'    - t1 = time to peak
+#'    - r = shape parameter
+#'    - alpha = decay rate
 #' @param prec_hyp_param A [numeric] [vector] of 5 values corresponding to
 #' hyperprior diagonal entries for the precision matrix (i.e. inverse variance)
 #' representing prior covariance of uncertainty around `mu_hyp_param`.
-#' If specified, must be 5 values long:
-#'    - defaults: y0 = 1.0, y1 = 0.00001, t1 = 1.0, r = 0.001, alpha = 1.0
+#' Must be 5 values long corresponding to the 5 estimated parameters.
 #' @param omega_param A [numeric] [vector] of 5 values corresponding to the
 #' diagonal entries representing the Wishart hyperprior
 #' distributions of `prec_hyp_param`, describing how much we expect parameters
 #' to vary between individuals.
-#' If specified, must be 5 values long:
-#'    - defaults: y0 = 1.0, y1 = 50.0, t1 = 1.0, r = 10.0, alpha = 1.0
+#' Must be 5 values long corresponding to the 5 estimated parameters.
 #' @param wishdf_param An [integer] [vector] of 1 value specifying the degrees
 #' of freedom for the Wishart hyperprior distribution of `prec_hyp_param`.
-#' If specified, must be 1 value long.
-#'    - default = 20.0
+#' Must be 1 value long.
 #'    - The value of `wishdf_param` controls how informative the Wishart prior
 #'      is. Higher values lead to tighter priors on individual variation.
 #'      Lower values (e.g., 5–10) make the prior more weakly informative,
@@ -36,8 +33,7 @@
 #' @param prec_logy_hyp_param A [numeric] [vector] of 2 values corresponding to
 #' hyperprior diagonal entries on the log-scale for the precision matrix
 #' (i.e. inverse variance) representing prior beliefs of individual variation.
-#' If specified, must be 2 values long:
-#'    - defaults = 4.0, 1.0
+#' Must be 2 values long.
 #'
 #' @returns A "curve_params_priors" object 
 #' (a subclass of [list] with the inputs to `prep_priors()` attached 
@@ -65,11 +61,11 @@
 #' @example inst/examples/examples-prep_priors.R
 
 prep_priors <- function(max_antigens,
-                        mu_hyp_param = c(1.0, 7.0, 1.0, -4.0, -1.0),
-                        prec_hyp_param = c(1.0, 0.00001, 1.0, 0.001, 1.0),
-                        omega_param = c(1.0, 50.0, 1.0, 10.0, 1.0),
-                        wishdf_param = 20,
-                        prec_logy_hyp_param = c(4.0, 1.0)) {
+                        mu_hyp_param = NULL,
+                        prec_hyp_param = NULL,
+                        omega_param = NULL,
+                        wishdf_param = NULL,
+                        prec_logy_hyp_param = NULL) {
 
   # Ensuring the length of specified priors is correct.
   # mu_hyp_param
--- a/man/prep_priors.Rd
+++ b/man/prep_priors.Rd
@@ -6,11 +6,11 @@
 \usage{
 prep_priors(
   max_antigens,
-  mu_hyp_param = c(1, 7, 1, -4, -1),
-  prec_hyp_param = c(1, 1e-05, 1, 0.001, 1),
-  omega_param = c(1, 50, 1, 10, 1),
-  wishdf_param = 20,
-  prec_logy_hyp_param = c(4, 1)
+  mu_hyp_param = NULL,
+  prec_hyp_param = NULL,
+  omega_param = NULL,
+  wishdf_param = NULL,
+  prec_logy_hyp_param = NULL
 )
 }
 \arguments{
@@ -20,37 +20,30 @@ antigen-isotypes (biomarkers) will be modeled.}
 \item{mu_hyp_param}{A \link{numeric} \link{vector} of 5 values representing the prior
 mean for the population level parameters
 parameters (y0, y1, t1, r, alpha) for each biomarker.
-If specified, must be 5 values long, representing the following parameters:
+Must be 5 values long, representing the following parameters:
 \itemize{
-\item y0 = baseline antibody concentration (default = 1.0)
-\item y1 = peak antibody concentration (default = 7.0)
-\item t1 = time to peak (default = 1.0)
-\item r = shape parameter (default = -4.0)
-\item alpha = decay rate (default = -1.0)
+\item y0 = baseline antibody concentration
+\item y1 = peak antibody concentration
+\item t1 = time to peak
+\item r = shape parameter
+\item alpha = decay rate
 }}
 
 \item{prec_hyp_param}{A \link{numeric} \link{vector} of 5 values corresponding to
 hyperprior diagonal entries for the precision matrix (i.e. inverse variance)
 representing prior covariance of uncertainty around \code{mu_hyp_param}.
-If specified, must be 5 values long:
-\itemize{
-\item defaults: y0 = 1.0, y1 = 0.00001, t1 = 1.0, r = 0.001, alpha = 1.0
-}}
+Must be 5 values long corresponding to the 5 estimated parameters.}
 
 \item{omega_param}{A \link{numeric} \link{vector} of 5 values corresponding to the
 diagonal entries representing the Wishart hyperprior
 distributions of \code{prec_hyp_param}, describing how much we expect parameters
 to vary between individuals.
-If specified, must be 5 values long:
-\itemize{
-\item defaults: y0 = 1.0, y1 = 50.0, t1 = 1.0, r = 10.0, alpha = 1.0
-}}
+Must be 5 values long corresponding to the 5 estimated parameters.}
 
 \item{wishdf_param}{An \link{integer} \link{vector} of 1 value specifying the degrees
 of freedom for the Wishart hyperprior distribution of \code{prec_hyp_param}.
-If specified, must be 1 value long.
+Must be 1 value long.
 \itemize{
-\item default = 20.0
 \item The value of \code{wishdf_param} controls how informative the Wishart prior
 is. Higher values lead to tighter priors on individual variation.
 Lower values (e.g., 5–10) make the prior more weakly informative,
@@ -60,10 +53,7 @@ which can help improve convergence if the model is over-regularized.
 \item{prec_logy_hyp_param}{A \link{numeric} \link{vector} of 2 values corresponding to
 hyperprior diagonal entries on the log-scale for the precision matrix
 (i.e. inverse variance) representing prior beliefs of individual variation.
-If specified, must be 2 values long:
-\itemize{
-\item defaults = 4.0, 1.0
-}}
+Must be 2 values long.}
 }
 \value{
 A "curve_params_priors" object
--- a/tests/testthat/test-run_serodynamics.R
+++ b/tests/testthat/test-run_serodynamics.R
@@ -27,7 +27,12 @@ test_that(
       nmc = 10,
       niter = 10, # Number of iterations
       strat = "strat", # Variable to be stratified
-      with_pop_params = TRUE
+      with_pop_params = TRUE,
+      mu_hyp_param = c(1.0, 7.0, 1.0, -4.0, -1.0),
+      prec_hyp_param = c(1.0, 0.00001, 1.0, 0.001, 1.0),
+      omega_param = c(1.0, 50.0, 1.0, 10.0, 1.0),
+      wishdf_param = 20,
+      prec_logy_hyp_param = c(4.0, 1.0)
     ) |>
       suppressWarnings()
 
@@ -95,7 +100,12 @@ test_that(
       strat = "bldculres", # Variable to be stratified by
       with_post = TRUE,
       with_pop_params = TRUE,
-      preclogy_per_iso = TRUE
+      preclogy_per_iso = TRUE,
+      mu_hyp_param = c(1.0, 7.0, 1.0, -4.0, -1.0),
+      prec_hyp_param = c(1.0, 0.00001, 1.0, 0.001, 1.0),
+      omega_param = c(1.0, 50.0, 1.0, 10.0, 1.0),
+      wishdf_param = 20,
+      prec_logy_hyp_param = c(4.0, 1.0)
     ) |>
       suppressWarnings()
 
@@ -204,7 +214,12 @@ test_that(
       strat = "bldculres", # Variable to be stratified by
       with_post = TRUE,
       with_pop_params = TRUE,
-      preclogy_per_iso = TRUE
+      preclogy_per_iso = TRUE,
+      mu_hyp_param = c(1.0, 7.0, 1.0, -4.0, -1.0),
+      prec_hyp_param = c(1.0, 0.00001, 1.0, 0.001, 1.0),
+      omega_param = c(1.0, 50.0, 1.0, 10.0, 1.0),
+      wishdf_param = 20,
+      prec_logy_hyp_param = c(4.0, 1.0)
     ) |>
       suppressWarnings()
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -78,5 +78,4 @@ Depends:
     R (>= 4.1.0)
 LazyData: true
 VignetteBuilder: knitr, quarto
-Config/roxygen2/version: 8.0.0
-RoxygenNote: 8.0.0
+Config/roxygen2/version: 8.1.0
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: serodynamics
 Title: Modeling Longitudinal Antibody Responses to Infection
-Version: 0.1.0.9013
+Version: 0.1.0.9014
 Authors@R: c(
     person("Peter", "Teunis", , "p.teunis@emory.edu", role = c("aut", "cph"),
            comment = c("Author of the method and original code.",
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -23,30 +23,40 @@ export(serodynamics_example)
 export(sim_case_data)
 export(sim_n_obs)
 export(summarize_posterior)
-importFrom(dplyr,.data)
-importFrom(dplyr,all_of)
-importFrom(dplyr,filter)
-importFrom(dplyr,mutate)
-importFrom(dplyr,n)
-importFrom(dplyr,reframe)
-importFrom(dplyr,rename)
-importFrom(dplyr,rowwise)
-importFrom(dplyr,select)
-importFrom(dplyr,ungroup)
+importFrom(dplyr,
+  .data,
+  all_of,
+  filter,
+  mutate,
+  n,
+  reframe,
+  rename,
+  rowwise,
+  select,
+  ungroup
+)
 importFrom(ggmcmc,ggs)
-importFrom(ggplot2,autoplot)
-importFrom(ggplot2,theme_bw)
-importFrom(ggplot2,vars)
-importFrom(rlang,.data)
-importFrom(rlang,.env)
+importFrom(ggplot2,
+  autoplot,
+  theme_bw,
+  vars
+)
+importFrom(rlang,
+  .data,
+  .env
+)
 importFrom(runjags,run.jags)
-importFrom(serocalculator,get_biomarker_levels)
-importFrom(serocalculator,get_biomarker_names)
-importFrom(serocalculator,get_biomarker_names_var)
-importFrom(serocalculator,get_values)
-importFrom(serocalculator,ids)
-importFrom(serocalculator,ids_varname)
-importFrom(stats,complete.cases)
-importFrom(stats,quantile)
+importFrom(serocalculator,
+  get_biomarker_levels,
+  get_biomarker_names,
+  get_biomarker_names_var,
+  get_values,
+  ids,
+  ids_varname
+)
+importFrom(stats,
+  complete.cases,
+  quantile
+)
 importFrom(tidyr,pivot_wider)
 importFrom(utils,read.csv)
--- a/R/run_serodynamics.R
+++ b/R/run_serodynamics.R
@@ -5,7 +5,9 @@
 #'  [runjags::run.jags()] as an MCMC
 #'  Bayesian model to estimate antibody dynamic curve parameters.
 #'  The [rjags::jags.model()] models seroresponse dynamics to an
-#'  infection. The antibody dynamic curve includes the following parameters:
+#'  infection. Priors are required for modeling and must be specified prior to
+#'  running `run_serodynamics`. The antibody dynamic curve includes the 
+#'  following parameters:
 #'  - y0 = baseline antibody concentration
 #'  - y1 = peak antibody concentration
 #'  - t1 = time to peak
--- a/man/run_serodynamics.Rd
+++ b/man/run_serodynamics.Rd
@@ -80,34 +80,27 @@ antigen-isotypes (biomarkers) will be modeled.}
     \item{\code{mu_hyp_param}}{A \link{numeric} \link{vector} of 5 values representing the prior
 mean for the population level parameters
 parameters (y0, y1, t1, r, alpha) for each biomarker.
-If specified, must be 5 values long, representing the following parameters:
+Must be 5 values long, representing the following parameters:
 \itemize{
-\item y0 = baseline antibody concentration (default = 1.0)
-\item y1 = peak antibody concentration (default = 7.0)
-\item t1 = time to peak (default = 1.0)
-\item r = shape parameter (default = -4.0)
-\item alpha = decay rate (default = -1.0)
+\item y0 = baseline antibody concentration
+\item y1 = peak antibody concentration
+\item t1 = time to peak
+\item r = shape parameter
+\item alpha = decay rate
 }}
     \item{\code{prec_hyp_param}}{A \link{numeric} \link{vector} of 5 values corresponding to
 hyperprior diagonal entries for the precision matrix (i.e. inverse variance)
 representing prior covariance of uncertainty around \code{mu_hyp_param}.
-If specified, must be 5 values long:
-\itemize{
-\item defaults: y0 = 1.0, y1 = 0.00001, t1 = 1.0, r = 0.001, alpha = 1.0
-}}
+Must be 5 values long corresponding to the 5 estimated parameters.}
     \item{\code{omega_param}}{A \link{numeric} \link{vector} of 5 values corresponding to the
 diagonal entries representing the Wishart hyperprior
 distributions of \code{prec_hyp_param}, describing how much we expect parameters
 to vary between individuals.
-If specified, must be 5 values long:
-\itemize{
-\item defaults: y0 = 1.0, y1 = 50.0, t1 = 1.0, r = 10.0, alpha = 1.0
-}}
+Must be 5 values long corresponding to the 5 estimated parameters.}
     \item{\code{wishdf_param}}{An \link{integer} \link{vector} of 1 value specifying the degrees
 of freedom for the Wishart hyperprior distribution of \code{prec_hyp_param}.
-If specified, must be 1 value long.
+Must be 1 value long.
 \itemize{
-\item default = 20.0
 \item The value of \code{wishdf_param} controls how informative the Wishart prior
 is. Higher values lead to tighter priors on individual variation.
 Lower values (e.g., 5–10) make the prior more weakly informative,
@@ -116,10 +109,7 @@ which can help improve convergence if the model is over-regularized.
     \item{\code{prec_logy_hyp_param}}{A \link{numeric} \link{vector} of 2 values corresponding to
 hyperprior diagonal entries on the log-scale for the precision matrix
 (i.e. inverse variance) representing prior beliefs of individual variation.
-If specified, must be 2 values long:
-\itemize{
-\item defaults = 4.0, 1.0
-}}
+Must be 2 values long.}
   }}
 }
 \value{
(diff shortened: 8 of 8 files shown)


sschildhauer:
@claude, review

sschildhauer:
@claude, review

R/prep_priors.R
```diff
@@ -1,43 +1,39 @@
...
-#'    - y1 = peak antibody concentration (default = 7.0)
-#'    - t1 = time to peak (default = 1.0)
-#'    - r = shape parameter (default = -4.0)
-#'    - alpha = decay rate (default = -1.0)
+#' Must be 5 values long, representing the following parameters:
+#'    - y0 = baseline antibody concentration
+#'    - y1 = peak antibody concentration
+#'    - t1 = time to peak
+#'    - r = shape parameter
+#'    - alpha = decay rate
```
imelainew: `configure_decay_priors()` `R/decay_helpers.R:43-54` drops the 5th element of `mu_hyp_param`, `prec_hyp_param`, and `omega_param` before they reach JAGS. So with **exponential decay**, users now have to supply a `r` `(shape`) prior that's silently discarded, which was invisible when defaults existed, but is now something they actively choose and don't get.

Would it be worth adding a `cli::cli_inform()` in `configure_decay_priors()` noting the 5th prior isn't used? (Accepting 4-element vectors would be cleaner but needs `decay_type` threaded into `prep_priors()`) Not blocking and happy to defer.

tests/testthat/test-prep_priors.R
```diff
@@ -13,3 +8,17 @@ test_that("priors are modifiable", {
...
+    expect_error()
+})
+
+test_that("Expect error when only some supplied", {
+  prep_priors(max_antigens = 2, 
+              mu_hyp_param = c(1.0,  5.0, 0.0, -2.0, -3.0),
+              omega_param = c(1.0, 50.0, 1.0, 5.0, 1.0),
+              wishdf_param = 15,
+              prec_logy_hyp_param = c(4.0, 1.0)) |>
+    expect_error()
```
imelainew: A question rather than a correction, I couldn't run R to check.
If we reverted `R/prep_priors.R:76` to the duplicated `length(mu_hyp_param`) check, this call would pass all five length checks (`mu_hyp_param is valid`) and then error anyway at `prec_hyp[k.test, , ] <- diag(NULL)` on zero-length replacement. If so, the bare `expect_error()` passes under the old bug too.

Would `expect_error`(`"Need to specify 5 priors for.*prec_hyp_param"`) pin it down? Same idea for line 14. Might be worth a quick local revert to confirm first.

vignettes/articles/getting-started.qmd
```diff
@@ -139,7 +139,12 @@ fitted_model <- run_serodynamics(
   nadapt = 100,    # Adaptation iterations
   nburn = 100,     # Burn-in iterations  
   nmc = 10,        # Samples per chain (use 1000+ for real analysis)
-  niter = 20       # Total iterations (use 2000+ for real analysis)
+  niter = 20,       # Total iterations (use 2000+ for real analysis)
+  mu_hyp_param = c(1.0, 7.0, 1.0, -4.0, -1.0), # Prior mean for the population level parameters
+  prec_hyp_param = c(1.0, 0.00001, 1.0, 0.001, 1.0), # Hyperprior precision matrix
+  omega_param = c(1.0, 50.0, 1.0, 10.0, 1.0), # Wishart hyperprior distributions
+  wishdf_param = 20, # Degrees of freedom for the Wishart hyperprior distribution
+  prec_logy_hyp_param = c(4.0, 1.0) # Hyperprior  precision matrix
```
imelainew, line 195: These are exactly the old defaults. That's the right choice for keeping the vignette output stable, but it means the getting-started doc now hands users five numbers to copy without saying where they came from or how to adjust them, which recreates the default, just in the docs instead of the code.

Would a sentence above this chunk help? Something like "these values are reasonable starting points for enteric fever serology (see Teunis et al. 2016); adjust based on your biomarker and expected kinetics." Non-blocking, but it seems close to the point of removing the defaults.

imelainew (commented):
Three non-blocking things below. The one on `test-prep_priors.R` is a question rather than a correction. If that test does hold up, I'm happy to approve.

R/prep_priors.R
sschildhauer: acknowledged in prep_priors for cli_abort options passing decay_type through

tests/testthat/test-prep_priors.R
sschildhauer: This line was meant to test if only one prior is missing will the test still error. I wasn't writing it in response to the old bug. But, I think it is better to be more specific. I added a piece saying what the expected error might be.

vignettes/articles/getting-started.qmd
sschildhauer: I think this is an ongoing discussion and something we might want to change in the future. As you said, these were taken from a typhoid analysis that Dr. Teunis was conducting, but I think we may want to ask Ezra and Kristen their thoughts on what priors we should have user-facing. If this passes, I can create an issue to further explore what these priors should be.

imelainew: Sounds good, agreed this belongs in a separate discussion with Ezra and Kristen. Could you link the issue here once you create it? Happy to resolve this thread then.

R/prep_priors.R
```diff
@@ -65,24 +66,34 @@
...
-                        wishdf_param = 20,
-                        prec_logy_hyp_param = c(4.0, 1.0)) {
+                        mu_hyp_param = NULL,
+                        prec_hyp_param = NULL,
+                        omega_param = NULL,
+                        wishdf_param = NULL,
+                        prec_logy_hyp_param = NULL,
+                        decay_type = "power") {
 
   # Ensuring the length of specified priors is correct.
```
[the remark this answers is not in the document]
sschildhauer: Added match.arg but kept default as power because we don't anticipate people using exponential close to as much

R/prep_priors.R
```diff
@@ -1,43 +1,44 @@
...
-#' parameters (y0, y1, t1, r, alpha) for each biomarker.
-#' If specified, must be 5 values long, representing the following parameters:
-#'    - y0 = baseline antibody concentration (default = 1.0)
-#'    - y1 = peak antibody concentration (default = 7.0)
-#'    - t1 = time to peak (default = 1.0)
-#'    - r = shape parameter (default = -4.0)
-#'    - alpha = decay rate (default = -1.0)
+#' parameters (y0, y1, t1, r, alpha) for each biomarker. When `
+#' decay_type == "exponential"` only 5 parameters (y0, y1, t1, alpha) are used.
+#' Must be 5 values long, representing the following parameters:
```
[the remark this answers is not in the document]
sschildhauer: Fixed the documentation as directed

R/prep_priors.R
```diff
@@ -65,24 +66,34 @@
...
+                        decay_type = "power") {
 
   # Ensuring the length of specified priors is correct.
   # mu_hyp_param
-  if (length(mu_hyp_param) != 5) {
+  if (length(mu_hyp_param) != 5 && decay_type == "power") {
     cli::cli_abort("Need to specify 5 priors for {.arg mu_hyp_param}")
+  } else if (length(mu_hyp_param) != 4 && decay_type == "exponential") {
+    cli::cli_abort("Need to specify 4 priors for {.arg mu_hyp_param}
+                   when decay_type == exponential")
```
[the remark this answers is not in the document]
sschildhauer: Added .arg decay type across cli_abort calls

R/prep_priors.R
```diff
@@ -95,7 +106,9 @@ prep_priors <- function(max_antigens,
 
 
   # Model parameters
-  n_params <- 5 # Assuming 5 model parameters [ y0, y1, t1, alpha, shape]
+  n_params <- ifelse(decay_type == "power", 5, 4) 
+  # Assuming 5 model parameters [ y0, y1, t1, alpha, shape], 4 if exponential
```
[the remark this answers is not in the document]
sschildhauer: Implemented suggestion

NEWS.md
```diff
@@ -74,6 +73,9 @@
 * Added a `CLAUDE.md` review-guideline item flagging roxygen doc copy-paste (use `@inheritParams`/`@inheritDotParams`/`@inheritSection` instead) and manual argument relaying (use `...` passthrough instead) (closes #262).
 
 ## New features
+* Default priors are no longer supplied for`run_serodynamics()`.  
+  **Breaking change:** Users must manually specify priors to run function or an
+  error will occur.
```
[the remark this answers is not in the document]
sschildhauer: Added recommendation

man/prep_priors.Rd
```diff
@@ -60,10 +57,19 @@ which can help improve convergence if the model is over-regularized.
...
+in the model. Options are \code{"power"} and \code{"exponential"}. Default is
+\code{"power"}. The \code{"power"} option uses
+\code{y(t) = (y1^(1-shape) - (1-shape)*alpha*(t-t1))^(1/(1-shape))}. The
+\code{"exponential"} option uses \code{y(t) = y1 * exp(-alpha*(t-t1))}.
+The exponential model does not estimate \code{shape}; its processed output
+includes \code{shape = 1} as a fixed value to preserve the common output
+structure.
+Note: \code{prep_priors()} still validates 5-element prior vectors. For
+exponential decay, the fifth (\code{shape}) prior is required for validation
+but is ignored and omitted from the reported priors.}
```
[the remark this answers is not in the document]
sschildhauer: Corrected in run_serodynamics decay_type documentation

R/prep_priors.R
```diff
@@ -65,24 +66,34 @@
...
-                        prec_hyp_param = c(1.0, 0.00001, 1.0, 0.001, 1.0),
-                        omega_param = c(1.0, 50.0, 1.0, 10.0, 1.0),
-                        wishdf_param = 20,
-                        prec_logy_hyp_param = c(4.0, 1.0)) {
+                        mu_hyp_param = NULL,
+                        prec_hyp_param = NULL,
+                        omega_param = NULL,
+                        wishdf_param = NULL,
+                        prec_logy_hyp_param = NULL,
+                        decay_type = "power") {
```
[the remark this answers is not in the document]
sschildhauer: Implemented suggestion

tests/testthat/test-prep_priors.R
```diff
@@ -13,3 +8,30 @@ test_that("priors are modifiable", {
...
+              wishdf_param = 15,
+              prec_logy_hyp_param = c(4.0, 1.0),
+              decay_type = "exponential") |>
+    expect_snapshot_value(style = "deparse")
+})
+
+test_that("Expect error when priors are not supplied", {
+  prep_priors(max_antigens = 2) |>
+    expect_error()
+})
```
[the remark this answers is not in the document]
sschildhauer: Added in explicit error messages

sschildhauer:
@claude, please review and recommend changes regarding the failing R-CMD-check from windows.

imelainew (approved):
Looks good! Ready for Ezra from my side.

sschildhauer:
Hi @d-morrison, this PR is ready for your review. It is not passing the windows R-CMD check, but it looks like it is a rjags installation issue that may require workflow changes, which I am hesitant to jump into myself. Thank you for taking a look.

R/prep_priors.R
```diff
@@ -95,7 +106,9 @@ prep_priors <- function(max_antigens,
 
 
   # Model parameters
-  n_params <- 5 # Assuming 5 model parameters [ y0, y1, t1, alpha, shape]
+  n_params <- if (decay_type == "power") 5L else 4L
```
imelainew, line 116: I have a question about order right: y0, y1, t1, alpha, shape, and it matches model.jags where position 4 is alpha and position 5 is log(shape minus 1). But the roxygen for mu_hyp_param above lists it as y0, y1, t1, r, alpha, which swaps the last two, and then the sentence below that says exponential uses y0, y1, t1, alpha, contradicting its own list. That ordering was already wrong on main, but it matters more now that exponential requires exactly 4 values: someone following the roxygen list will pass the shape prior as their fourth value and it will be used as the prior on log alpha with no error. Worth fixing the roxygen to match this line while you're in here. 
Two smaller things in the same block: the exponential sentence says 5 parameters where it should say 4, and the returns section still describes n_params as 5 and the array dimensions as max_antigens by 5.

d-morrison:
> Hi @d-morrison, this PR is ready for your review. It is not passing the windows R-CMD check, but it looks like it is a rjags installation issue that may require workflow changes, which I am hesitant to jump into myself. Thank you for taking a look.

@sschildhauer fixed!

d-morrison:
> **Claude finished review** — [View run](https://github.com/UCD-SERG/serodynamics/actions/runs/33375496998)
> 
> Confirmed. Now I have everything needed to write the review.
> 
> ## Review: #284 — "No default priors"
> Follow-up review. Since the [prior round](https://github.com/UCD-SERG/serodynamics/actions/runs/32511935304) (commit `4908763`), the author pushed the commit range `49087638..e2157ae2` (24 commits), which primarily adds exponential-decay-aware prior validation to `prep_priors()` and hardens the CI JAGS-install steps. I re-verified the file contents directly and confirmed the previously-open finding.
> 
> ### Previously-flagged issues: status
> **#2 — `vignettes/articles/getting-started.qmd` stratified-example call missing priors — Fixed.** `vignettes/articles/getting-started.qmd:220-234` now supplies all five explicit prior arguments to the `fitted_stratified` call, matching the fix already applied to the earlier `run-model` chunk. Both `run_serodynamics()` calls in the vignette now have explicit priors, so the vignette should render end-to-end again.
> 
> **#6 (remaining half) — NEWS.md wording/labeling.** `NEWS.md:5-7` now carries an explicit `**Breaking change:**` label (matching the `NEWS.md:104`-style precedent) and the earlier missing-space typo before the backtick is fixed. This closes out both outstanding sub-points from the prior round.
> 
> ### New findings (introduced in `49087638..e2157ae2`)
> **1. (Blocking) `prep_priors()` silently accepts an invalid `decay_type`, bypassing all length validation — `R/prep_priors.R:78-96,109`** The validation for `mu_hyp_param`/`prec_hyp_param`/`omega_param` is now written as two mutually exclusive branches, gated on `decay_type == "power"` and `decay_type == "exponential"` respectively — with no `else`/default branch and no `match.arg()`/`arg.match()` call on `decay_type` itself:
> 
> ```r
> if (length(mu_hyp_param) != 5 && decay_type == "power") {
>   cli::cli_abort(...)
> } else if (length(mu_hyp_param) != 4 && decay_type == "exponential") {
>   cli::cli_abort(...)
> }
> ```
> 
> and later:
> 
> ```r
> n_params <- if (decay_type == "power") 5L else 4L
> ```
> 
> If `decay_type` is anything other than the exact strings `"power"`/`"exponential"` (a typo, wrong case, trailing whitespace, `NULL`, etc.), **neither** `cli_abort()` branch fires regardless of the supplied prior lengths — the length checks are simply skipped — and `n_params` falls through to `4L` (the "exponential" branch) via the bare `else`. A caller who does `prep_priors(max_antigens = 2, mu_hyp_param = 1:5, ..., decay_type = "Power")` gets no error at all and silently gets treated as a 4-parameter exponential model.
> 
> This is not hypothetical: `prep_priors()` is `@export`ed and documented/called directly (e.g. `inst/examples/examples-prep_priors.R`, `tests/testthat/fixtures/make-example-runjags-inputs.R`), so it is reachable by a user without going through `run_serodynamics()`'s `match.arg(decay_type, c("power", "exponential"))` guard (`R/run_serodynamics.R:129`), which is the only place `decay_type` is actually validated in this PR. The git history shows this was previously caught: commit `5effb7b` ("Adding match.arg for decay_type") added exactly this guard directly inside `prep_priors()`, but the very next relevant commit, `a280916` ("Changing workflow so run_serodynamics explicitly passes decay_type to prep_priors"), removed it again — leaving `prep_priors()` unguarded. This is the same class of bug as finding #1 from the very first review round (a validation branch that silently passes through instead of erroring), and it directly undermines this PR's stated goal of forcing users to supply correct, deliberate priors: an invalid `decay_type` typo combined with a 5-length `mu_hyp_param` now runs to completion with a mismatched, half-specified model instead of failing loudly. Suggest restoring `decay_type <- match.arg(decay_type, c("power", "exponential"))` as the first line of the function body (as commit `5effb7b` had it).
> 
> There is also no test covering this path — `tests/testthat/test-prep_priors.R` exercises `"power"` and `"exponential"` explicitly, but nothing exercises an invalid/misspelled `decay_type`, which would have caught this.
> 
> **2. (Minor) Roxygen doc miscounts exponential parameters — `R/prep_priors.R:18-19`, propagated to `man/prep_priors.Rd:32` and `man/run_serodynamics.Rd:91`**
> 
> ```r
> #' When `decay_type = "exponential"` only 5 parameters (y0, y1, t1, alpha) are 
> #' used.
> ```
> 
> The sentence says "5 parameters" but lists four (`y0, y1, t1, alpha`) — for exponential decay it should read "4 parameters," consistent with the rest of the same paragraph (`prec_hyp_param`/`omega_param` docs correctly say "4 values ... when `decay_type = "exponential"`") and with the validation code itself (`length(...) != 4 && decay_type == "exponential"`). This propagates automatically into both `.Rd` files since `run_serodynamics.Rd`'s copy is roxygen-inherited (`@inheritDotParams prep_priors`), so a single one-word fix in `R/prep_priors.R` resolves all three occurrences.
> 
> ### Not flagged (checked, no issue)
> * `@inheritParams run_serodynamics decay_type` (`R/prep_priors.R:41`) is valid roxygen2 syntax — `@inheritParams` accepts an optional list of specific argument names after the source function (same mechanism as `@inheritDotParams`), confirmed against the roxygen2 `tags-reuse` reference docs. `man/prep_priors.Rd`'s `decay_type` entry is correctly generated from it, not hand copy-pasted.
> * The "Aiemjoy K et al. ... (2022)" citation added to `vignettes/articles/getting-started.qmd:143` is a real, verifiable paper (_Lancet Microbe_ 2022;3:e578-87) — not a hallucination.
> * New tests `"Preparing priors for exponential"`, `"Expect error for {mu_hyp_param,prec_hyp_param,omega_param} under exponential"` correctly exercise the exponential branch and its error paths; snapshot additions in `tests/testthat/_snaps/prep_priors.md` match.
> * `test-plot_residuals.R`'s new heavy-JAGS exponential test correctly routes through `run_serodynamics(decay_type = "exponential", ...)`, which is protected by `match.arg()`, so it isn't exposed to finding [Jags2 #1](https://github.com/UCD-SERG/serodynamics/pull/1).
> * `.github/workflows/*.yaml` changes (from the saved diff, since the on-disk copies are the default-branch fallback per this run's instructions): the Windows JAGS install fix (`start /wait "" wjags.exe /S` instead of a bare, non-blocking `wjags.exe /S`) is a genuine correctness fix — previously the script could delete/move on before the installer finished. The added `curl -fSL --retry 3 --retry-connrefused` flags are a reasonable hardening of a flaky download. The `install rjags` step's switch to `if (!requireNamespace("rjags", quietly = TRUE)) install.packages(...)` and dropping `type = "source"` is a plausible optimization (CRAN's rjags binaries for Windows/macOS are built against a specific JAGS version, and the workflow pins the same JAGS 4.3.1 that's installed), but I can't fully verify from a static read that binary rjags will always link correctly against the separately-installed JAGS across all matrix legs — flagging as informational rather than blocking since I can't execute the workflow to confirm.
> * No `.data`-in-tidyselect, branching-that-only-varies-columns, `merge()`-vs-join, or missing `relationship=` issues — this diff introduces no new dplyr code (the one dplyr-adjacent test-file change, `test-plot_residuals.R:99`, is a pure whitespace fix, `attr(model,"decay_type")` → `attr(model, "decay_type")`, functionally identical).
> * `DESCRIPTION` version bump (`.9017` → `.9018`) is a standard sequential dev bump.
> 
> ### Verdict
> **Needs more work** — finding #1 (`prep_priors()` no longer validates `decay_type`, silently falling through to the 4-parameter/exponential path for any unrecognized value) is a genuine regression introduced in this round and undercuts the PR's core goal of forcing users to supply correct, deliberate priors. Finding #2 is a one-word doc fix. All previously-open findings from earlier rounds (the vignette's second `run_serodynamics()` call, and the NEWS.md wording/label) are now resolved.
> 
> Reviewed commit: [e2157ae](https://github.com/UCD-SERG/serodynamics/commit/e2157ae2c2349538364b44bed251449236d00c54)

@sschildhauer please address this review

sschildhauer:
> > **Claude finished review** — [View run](https://github.com/UCD-SERG/serodynamics/actions/runs/33375496998)
> > Confirmed. Now I have everything needed to write the review.
> > ## Review: #284 — "No default priors"
> > Follow-up review. Since the [prior round](https://github.com/UCD-SERG/serodynamics/actions/runs/32511935304) (commit `4908763`), the author pushed the commit range `49087638..e2157ae2` (24 commits), which primarily adds exponential-decay-aware prior validation to `prep_priors()` and hardens the CI JAGS-install steps. I re-verified the file contents directly and confirmed the previously-open finding.
> > ### Previously-flagged issues: status
> > **#2 — `vignettes/articles/getting-started.qmd` stratified-example call missing priors — Fixed.** `vignettes/articles/getting-started.qmd:220-234` now supplies all five explicit prior arguments to the `fitted_stratified` call, matching the fix already applied to the earlier `run-model` chunk. Both `run_serodynamics()` calls in the vignette now have explicit priors, so the vignette should render end-to-end again.
> > **#6 (remaining half) — NEWS.md wording/labeling.** `NEWS.md:5-7` now carries an explicit `**Breaking change:**` label (matching the `NEWS.md:104`-style precedent) and the earlier missing-space typo before the backtick is fixed. This closes out both outstanding sub-points from the prior round.
> > ### New findings (introduced in `49087638..e2157ae2`)
> > **1. (Blocking) `prep_priors()` silently accepts an invalid `decay_type`, bypassing all length validation — `R/prep_priors.R:78-96,109`** The validation for `mu_hyp_param`/`prec_hyp_param`/`omega_param` is now written as two mutually exclusive branches, gated on `decay_type == "power"` and `decay_type == "exponential"` respectively — with no `else`/default branch and no `match.arg()`/`arg.match()` call on `decay_type` itself:
> > ```r
> > if (length(mu_hyp_param) != 5 && decay_type == "power") {
> >   cli::cli_abort(...)
> > } else if (length(mu_hyp_param) != 4 && decay_type == "exponential") {
> >   cli::cli_abort(...)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and later:
> > ```r
> > n_params <- if (decay_type == "power") 5L else 4L
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If `decay_type` is anything other than the exact strings `"power"`/`"exponential"` (a typo, wrong case, trailing whitespace, `NULL`, etc.), **neither** `cli_abort()` branch fires regardless of the supplied prior lengths — the length checks are simply skipped — and `n_params` falls through to `4L` (the "exponential" branch) via the bare `else`. A caller who does `prep_priors(max_antigens = 2, mu_hyp_param = 1:5, ..., decay_type = "Power")` gets no error at all and silently gets treated as a 4-parameter exponential model.
> > This is not hypothetical: `prep_priors()` is `@export`ed and documented/called directly (e.g. `inst/examples/examples-prep_priors.R`, `tests/testthat/fixtures/make-example-runjags-inputs.R`), so it is reachable by a user without going through `run_serodynamics()`'s `match.arg(decay_type, c("power", "exponential"))` guard (`R/run_serodynamics.R:129`), which is the only place `decay_type` is actually validated in this PR. The git history shows this was previously caught: commit `5effb7b` ("Adding match.arg for decay_type") added exactly this guard directly inside `prep_priors()`, but the very next relevant commit, `a280916` ("Changing workflow so run_serodynamics explicitly passes decay_type to prep_priors"), removed it again — leaving `prep_priors()` unguarded. This is the same class of bug as finding #1 from the very first review round (a validation branch that silently passes through instead of erroring), and it directly undermines this PR's stated goal of forcing users to supply correct, deliberate priors: an invalid `decay_type` typo combined with a 5-length `mu_hyp_param` now runs to completion with a mismatched, half-specified model instead of failing loudly. Suggest restoring `decay_type <- match.arg(decay_type, c("power", "exponential"))` as the first line of the function body (as commit `5effb7b` had it).
> > There is also no test covering this path — `tests/testthat/test-prep_priors.R` exercises `"power"` and `"exponential"` explicitly, but nothing exercises an invalid/misspelled `decay_type`, which would have caught this.
> > **2. (Minor) Roxygen doc miscounts exponential parameters — `R/prep_priors.R:18-19`, propagated to `man/prep_priors.Rd:32` and `man/run_serodynamics.Rd:91`**
> > ```r
> > #' When `decay_type = "exponential"` only 5 parameters (y0, y1, t1, alpha) are 
> > #' used.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The sentence says "5 parameters" but lists four (`y0, y1, t1, alpha`) — for exponential decay it should read "4 parameters," consistent with the rest of the same paragraph (`prec_hyp_param`/`omega_param` docs correctly say "4 values ... when `decay_type = "exponential"`") and with the validation code itself (`length(...) != 4 && decay_type == "exponential"`). This propagates automatically into both `.Rd` files since `run_serodynamics.Rd`'s copy is roxygen-inherited (`@inheritDotParams prep_priors`), so a single one-word fix in `R/prep_priors.R` resolves all three occurrences.
> > ### Not flagged (checked, no issue)
> > 
> > * `@inheritParams run_serodynamics decay_type` (`R/prep_priors.R:41`) is valid roxygen2 syntax — `@inheritParams` accepts an optional list of specific argument names after the source function (same mechanism as `@inheritDotParams`), confirmed against the roxygen2 `tags-reuse` reference docs. `man/prep_priors.Rd`'s `decay_type` entry is correctly generated from it, not hand copy-pasted.
> > * The "Aiemjoy K et al. ... (2022)" citation added to `vignettes/articles/getting-started.qmd:143` is a real, verifiable paper (_Lancet Microbe_ 2022;3:e578-87) — not a hallucination.
> > * New tests `"Preparing priors for exponential"`, `"Expect error for {mu_hyp_param,prec_hyp_param,omega_param} under exponential"` correctly exercise the exponential branch and its error paths; snapshot additions in `tests/testthat/_snaps/prep_priors.md` match.
> > * `test-plot_residuals.R`'s new heavy-JAGS exponential test correctly routes through `run_serodynamics(decay_type = "exponential", ...)`, which is protected by `match.arg()`, so it isn't exposed to finding [Jags2 #1](https://github.com/UCD-SERG/serodynamics/pull/1).
> > * `.github/workflows/*.yaml` changes (from the saved diff, since the on-disk copies are the default-branch fallback per this run's instructions): the Windows JAGS install fix (`start /wait "" wjags.exe /S` instead of a bare, non-blocking `wjags.exe /S`) is a genuine correctness fix — previously the script could delete/move on before the installer finished. The added `curl -fSL --retry 3 --retry-connrefused` flags are a reasonable hardening of a flaky download. The `install rjags` step's switch to `if (!requireNamespace("rjags", quietly = TRUE)) install.packages(...)` and dropping `type = "source"` is a plausible optimization (CRAN's rjags binaries for Windows/macOS are built against a specific JAGS version, and the workflow pins the same JAGS 4.3.1 that's installed), but I can't fully verify from a static read that binary rjags will always link correctly against the separately-installed JAGS across all matrix legs — flagging as informational rather than blocking since I can't execute the workflow to confirm.
> > * No `.data`-in-tidyselect, branching-that-only-varies-columns, `merge()`-vs-join, or missing `relationship=` issues — this diff introduces no new dplyr code (the one dplyr-adjacent test-file change, `test-plot_residuals.R:99`, is a pure whitespace fix, `attr(model,"decay_type")` → `attr(model, "decay_type")`, functionally identical).
> > * `DESCRIPTION` version bump (`.9017` → `.9018`) is a standard sequential dev bump.
> > 
> > ### Verdict
> > **Needs more work** — finding #1 (`prep_priors()` no longer validates `decay_type`, silently falling through to the 4-parameter/exponential path for any unrecognized value) is a genuine regression introduced in this round and undercuts the PR's core goal of forcing users to supply correct, deliberate priors. Finding #2 is a one-word doc fix. All previously-open findings from earlier rounds (the vignette's second `run_serodynamics()` call, and the NEWS.md wording/label) are now resolved.
> > Reviewed commit: [e2157ae](https://github.com/UCD-SERG/serodynamics/commit/e2157ae2c2349538364b44bed251449236d00c54)
> 
> @sschildhauer please address this review

All issues addressed 
- Added additional cli_abort to test for spelling errors in decay type. This will be a good check overall for run_serodynamics as well. 
- Fixed mistake in prep_priors documentation.

sschildhauer:
@claude , review

d-morrison:
/review

Requesting a review on the current head `6553b87`. Automatic review is disabled in this repo (see #266), so nothing has reviewed this head on its own.

@sschildhauer — on your question about the failing Windows `R-CMD-check`: that was an `rjags` installation problem in CI rather than anything in this diff, and it has since been fixed on `main` by `ba4536c` ("harden JAGS download and rjags verification on Windows and macOS", merged via #308). The `windows-latest (release)` job on this PR's current head is now **green**, so that blocker is gone and no workflow changes are needed here.

The only remaining red is `codecov/project`, a coverage-percentage drop rather than a test failure.

Heads-up on merge order: this PR, #292, #298, #310 and #311 all carry `DESCRIPTION` version `0.1.0.9019`, so whichever merges second will conflict there and need a re-bump. Tracked as #312.

---

d-morrison:
Both round-3 findings **addressed** in `527da63`.

**1. `configure_decay_priors()` dead code — deleted.** Verified before removing:

```
$ grep -rn 'configure_decay_priors' --include='*.R' --include='*.Rd' --include='*.qmd' --include='NAMESPACE' .
./R/decay_helpers.R:38:configure_decay_priors <- function(priorspec, decay_type) {
```

One hit, its own definition. Not in `NAMESPACE`, no `man/` page, no test. `R/decay_helpers.R` goes 118 → 81 lines; the four surviving functions (`select_decay_model()`, `get_decay_monitors()`, `add_fixed_exponential_shape()`, `get_model_decay_type()`) are untouched.

**2. NEWS.md wording — qualified.** It now reads:

> `prep_priors` now takes 4 values for `mu_hyp_param`, `prec_hyp_param` and `omega_param` when `decay_type = "exponential"`, and 5 when it is `"power"`.

which no longer contradicts the `**Breaking change:**` sentence beneath it.

---

Two notes on the rest of the PR, neither requiring action here:

**#299 and #304 will close with this PR.** Both report `prep_priors()`'s second length check testing `mu_hyp_param` twice; this branch already fixes it. GitHub lists this PR as a closing reference for #299 but not #304 — worth adding `Closes #304` to the body so the duplicate doesn't linger. I left a triage note on #304 pointing here.

Worth knowing for the test #304 asks for: a wrong-length `prec_hyp_param` usually fails loudly downstream, but **length 1 fails silently**. `diag()` given a single number returns an identity matrix of that order, so `prec_hyp[k, , ] <- diag(5)` conforms and assigns cleanly — a user passing a scalar precision gets a silently different prior with no error. That is the one length no downstream error catches.

**PR #298 depends on this one merging first.** Its vignette says "No priors are specified by default", which is false against `main` and true once this PR lands. Its reviewer has flagged that as a blocking factual error across three rounds; I've rebutted it there as a merge-order constraint rather than a defect. Their file sets are disjoint, so nothing conflicts — but **#284 before #298**, or #298's author will rewrite correct prose.

---

d-morrison:
/review

---

d-morrison:
@sschildhauer — I pushed one commit to this branch that I should not have: `527da63`, which deleted `configure_decay_priors()` from `R/decay_helpers.R` and reworded a NEWS entry, both in response to the round-3 review. This PR is yours, not mine, and per @d-morrison I should only have been working PRs they opened or are assigned to.

Nothing here requires action from you. The commit is on your branch to keep or drop as you see fit; `git revert 527da63` restores the function if you want it back. I have stopped watching this PR and will not push to it again.

---

sschildhauer:
Hi @imelainew, requesting your review one more time on this PR before requesting from Ezra.

imelainew (approved):
Looks good!

sschildhauer:
Hi @d-morrison , this should be ready for review again. Thank you!

R/prep_priors.R
d-morrison: @imelainew @sschildhauer  is this question resolved? if so, please press the "resolve comment" button so I know not to worry about it

Image

vignettes/articles/getting-started.qmd
d-morrison: @sschildhauer @imelainew see above; please resolve all open threads

d-morrison (changes requested):
Looks good! Just resolve the remaining comment threads and I'll approve and merge.

R/prep_priors.R
sschildhauer: Fixed documentation to ensure that the order is correct.

vignettes/articles/getting-started.qmd
sschildhauer: I checked with Kristen and it seems okay to use these priors but add the reference to the paper that first produced them. I added this piece and instructed users to always specify their own priors. Aiemjoy K et al. Estimating typhoid incidence from community-based serosurveys: a multicohort study (2022)

sschildhauer:
@d-morrison , all comments are resolved and ready for your review. Thank you!

d-morrison (changes requested):
please remember to always get a fresh AI review on the latest commit before requesting human review; see https://github.com/UCD-SERG/serodynamics/pull/284#issuecomment-5630066512 which found one more issue

R/prep_priors.R
```diff
@@ -1,60 +1,65 @@
...
+#' Must be 5 values long corresponding to the 5 estimated parameters (4 values 
+#' when `decay_type = "exponential"`).
 #' @param omega_param A [numeric] [vector] of 5 values corresponding to the
 #' diagonal entries representing the Wishart hyperprior
 #' distributions of `prec_hyp_param`, describing how much we expect parameters
-#' to vary between individuals.
-#' If specified, must be 5 values long:
-#'    - defaults: y0 = 1.0, y1 = 50.0, t1 = 1.0, r = 10.0, alpha = 1.0
+#' to vary between individuals (4 values when `decay_type = "exponential"`).
+#' Must be 5 values long corresponding to the 5 estimated parameters.
```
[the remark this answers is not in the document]
sschildhauer: Fixed as requested

tests/testthat/test-run_serodynamics.R
```diff
@@ -205,7 +215,12 @@ test_that(
       strat = "bldculres", # Variable to be stratified by
       with_post = TRUE,
       with_pop_params = TRUE,
-      preclogy_per_iso = TRUE
+      preclogy_per_iso = TRUE,
+      mu_hyp_param = c(1.0, 7.0, 1.0, -4.0),
+      prec_hyp_param = c(1.0, 0.00001, 1.0, 0.001),
+      omega_param = c(1.0, 50.0, 1.0, 10.0),
```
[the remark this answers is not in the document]
sschildhauer: This test is calling decay_type = "exponential" and is working fine. This seems like a mistake.

R/prep_priors.R
```diff
@@ -1,60 +1,66 @@
...
 #' @param mu_hyp_param A [numeric] [vector] of 5 values representing the prior
 #' mean for the population level parameters
-#' parameters (y0, y1, t1, r, alpha) for each biomarker.
-#' If specified, must be 5 values long, representing the following parameters:
-#'    - y0 = baseline antibody concentration (default = 1.0)
-#'    - y1 = peak antibody concentration (default = 7.0)
-#'    - t1 = time to peak (default = 1.0)
-#'    - r = shape parameter (default = -4.0)
-#'    - alpha = decay rate (default = -1.0)
+#' parameters (y0, y1, t1, alpha, r) for each biomarker. Must be 5 values long
```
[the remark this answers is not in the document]
sschildhauer: Fixed wording to no longer require with "Must".

sschildhauer:
@claude, review

sschildhauer:
After a number of AI reviews I believe this is ready again for @d-morrison review.

d-morrison (approved):
LGTM - thanks!

Issue #277, opened 2026-07-28
dispatch-explicit-review double-dispatches reviews that the gha reusable already handles


d-morrison:
Follow-up from the review on #274 ([comment](https://github.com/UCD-SERG/serodynamics/pull/274#issuecomment-5101640221), finding 2). The reviewer flagged this at *moderate* confidence because it depends on behavior in the external reusable workflow. **Confirmed at high confidence** by reading that workflow directly — details below.

## The overlap is real

Two matchers can fire on the same comment, and neither knows about the other.

`d-morrison/gha`'s reusable `claude.yml` has its own review-dispatch step ("Dispatch code review on @claude review comment"), gated on a matcher that resolves to:

```bash
[[ "$COMMENT_BODY" =~ @claude[[:space:]]+review ]]
```

This repo's `dispatch-explicit-review` job matches:

```bash
[[ "$comment_body" =~ @claude[[:space:][:punct:]]*(please[[:space:]]+)?review([[:space:][:punct:]]|$) ]]
```

Comparing them:

| Comment | gha reusable | this repo's job | Result |
|---|---|---|---|
| `@claude review` | ✅ | ✅ | **two dispatches** |
| `@claude  review` (two spaces) | ✅ | ✅ | **two dispatches** |
| `@claude, please review` | ❌ | ✅ | one — the #230 case #274 fixed |
| `@claude can you review this?` | ❌ | ❌ | none — see #276 |

So the **plainest and most common phrasing**, `@claude review`, triggers two review runs, each spending `CLAUDE_CODE_OAUTH_TOKEN`. That isn't an edge case.

The reviewer also noted the local job has no `needs: claude`, so for a comment like `@claude review and fix the failing test` the two runs can review *different* heads — the local one fires immediately against the pre-fix head while the reusable's fires after Claude commits. One of the two reviews is then of stale code.

## Why the local job exists

Not redundancy for its own sake: the reusable's matcher requires whitespace and no punctuation between `@claude` and `review`, so it missed `@claude, please review` (#230). That gap is [`d-morrison/gha#315`](https://github.com/d-morrison/gha/issues/315), **now closed** — the fix replaced a literal `contains()` with the regex above, which tolerates *extra whitespace* but still not intervening punctuation or words.

## Options

[...]

PR #286, opened 2026-08-12
ci: accept `@claude review` again, and reply instead of going silent

### LINKED ISSUE #285: ci: `@claude review` is silently ignored -- the only review trigger is `/review`, which nobody uses
## Symptom

Commenting `@claude review` on a pull request does nothing at all. No review, no
acknowledgement, no failed check -- the workflow run appears and reports
`skipped` after about one second.

Most recently: [run 31573442943](https://github.com/UCD-SERG/serodynamics/actions/runs/31573442943),
triggered by `@claude review` on #230 at `2026-08-12T07:18:39Z`.

## Cause

#282 disabled the `@claude` agent (`claude.yml`) and moved review dispatch into
`claude-code-review.yml` behind a **new** trigger: a comment *starting with*
`/review`. The `@claude review` mention that everyone had been using was
dispatched by `claude.yml`, which is now switched off.

`dispatch-on-comment`'s gate is:

```yaml
startsWith(github.event.comment.body, '/review')
```

`@claude review` does not start with `/review`, so that job skips. The `review`
job below it is gated `github.event_name != 'issue_comment'`, so it skips too.
Both jobs skipping is why the run reports `skipped` rather than failing --
nothing is red, and nothing tells the commenter their request went nowhere.

## Scale

Every `issue_comment` run of `claude-code-review.yml` since the `/review` path
landed on 2026-07-31 has skipped -- 7 of 7. Four were ordinary or bot comments
and were correctly skipped. The other three were explicit review requests from
MEMBERs that got silence:

| When | Who | Comment | Run |
| --- | --- | --- | --- |
| 2026-07-31T17:53Z | @sschildhauer | `@claude, review` | [30652966747](https://github.com/UCD-SERG/serodynamics/actions/runs/30652966747) |
| 2026-08-11T18:21Z | @Kwan-Jenny | `@claude review` | [31522300736](https://github.com/UCD-SERG/serodynamics/actions/runs/31522300736) |
| 2026-08-12T07:18Z | @d-morrison | `@claude review` | [31573442943](https://github.com/UCD-SERG/serodynamics/actions/runs/31573442943) |

No one has typed `/review` even once in those 12 days. The trigger that exists
is the one nobody uses, and the one everybody uses does nothing.

## Why the original rationale does not apply here

`claude-code-review.yml`'s header explains the choice:

> It is a slash command rather than an `@claude ...` mention on purpose: the
> mention path lives in claude.yml, which is switched off.

Upstream's stub gives the underlying reason -- "any `@claude` also wakes
claude.yml", which would double-dispatch (the #277 incident). But `claude.yml`
is disabled in this repo, so there is no second path to collide with. The
constraint that motivated excluding the mention form is absent here, and
excluding it is what produces the silence.

## Proposed fix

Accept both forms in `dispatch-on-comment`:

- keep the existing standalone `/review` command, and
- add the `@claude review` mention, matched with
  `Morrison-Lab/gha/.github/actions/detect-review-request@v2` rather than a
  hand-rolled regex -- that is the canonical matcher, and it already
  distinguishes a review request from an agent task, ignores mentions quoted in
  code spans or blockquotes, and covers the `@claude, please review` phrasing.

Verified locally against the real comment bodies from this repo (10 cases, all
correct): `@claude review`, `@claude, review`, and `@claude, please review`
match; `@claude, please address the above issues ...`, `@copilot, please review
this PR`, a mention inside prose, one in a code span, and one in a blockquote do
not.

Also worth fixing the silence itself: reply when a trusted author mentions
`@claude` and it is *not* a review request, so "the agent is off" is visible
rather than inferred from nothing happening.

### Caution for whoever re-enables `claude.yml`

Adding the mention path here means re-enabling the agent would recreate #277 --
`@claude review` would match both `claude.yml`'s own matcher and this local
job, dispatching two paid review runs against possibly different heads. The
mention path must be removed from `claude-code-review.yml` at the same time the
agent is re-enabled.
</body>

d-morrison:
Fixes the skipped run at
https://github.com/UCD-SERG/serodynamics/actions/runs/31573442943.
Closes #285.

## What was wrong

`@claude review` has done nothing since #282. That PR disabled the agent and
moved review dispatch into `claude-code-review.yml` behind a comment
*starting with* `/review`, on the reasoning that the mention form belonged to
`claude.yml` -- which it had just switched off. So the mention stopped
dispatching anything.

It stopped *silently*, which is why it lasted. `dispatch-on-comment`'s gate
rejected the body, the `review` job skips every `issue_comment` run, and a run
whose jobs all skip reports `skipped` rather than failing. Nothing red, no
reply, nothing to notice.

Every review request in the twelve days since used the mention form and was
ignored. `/review` was never typed once:

| When | Who | Comment | Run |
| --- | --- | --- | --- |
| 2026-07-31T17:53Z | @sschildhauer | `@claude, review` | [30652966747](https://github.com/UCD-SERG/serodynamics/actions/runs/30652966747) |
| 2026-08-11T18:21Z | @Kwan-Jenny | `@claude review` | [31522300736](https://github.com/UCD-SERG/serodynamics/actions/runs/31522300736) |
| 2026-08-12T07:18Z | @d-morrison | `@claude review` | [31573442943](https://github.com/UCD-SERG/serodynamics/actions/runs/31573442943) |

Upstream's stub is `/review`-only because "any `@claude` also wakes
claude.yml" -- a collision that does not exist here, because claude.yml is off.
The constraint that motivated excluding the mention is absent, and excluding it
is what produced the silence.

## What changed

`dispatch-on-comment` accepts either spelling. The job-level `if:` widens to
any `@claude` mention as a pre-filter -- a GitHub Actions expression cannot
tell a review request from an agent task -- and the steps decide:

- **`detect-review-request@v2`** classifies the mention. Using gha's canonical
  matcher rather than a pattern of our own keeps the accepted phrasings and the
  code-span/blockquote stripping in one place.
- **A mention that is not a review request now gets a reply** saying the agent
  is off and naming the triggers that work, gated on **`detect-bot-mention@v2`**
  so a mention quoted in prose stays quiet. A skipped workflow is
  indistinguishable from a broken bot; that is the failure this PR is really
  about, and the mention path alone would leave it in place for
  `@claude, please fix X`.

Both matchers are `continue-on-error` and degrade in their own safe direction:
an unclassified mention still reaches the `/review` path, and an unclassified
reply is sent rather than withheld.

Bot-authored comments are excluded at the gate. The `author_association`
allowlist already covers them, but this job now quotes `@claude review` back
into the thread, so a hole in that allowlist would be a comment loop rather
than a stray run. Two independent guards, because the failure is expensive.

`claude.yml`'s re-enable instructions said "nothing else needs to change".
That is no longer true, and the header now says so: re-enabling the agent while
this mention path exists gives `@claude review` two dispatchers and reproduces
#277. Removing the mention path is now step three of re-enabling.

## Verification

No R code changed, so the R suite is not implicated; what needed testing was
the matcher and the branching, both exercised offline.

**The matcher**, run at `v2` against this repo's real comment bodies -- 10
cases, all correct. `@claude review`, `@claude, review` and
`@claude, please review` match; `@claude, please address the above issues ...`,
`@copilot, please review this PR`, a mention inside prose, one in a code span
and one in a blockquote do not.

One caveat worth recording: the matcher's stripper uses an awk interval
expression that **mawk cannot compile**, and on mawk it returns `false` for
every input with exit 0 -- a silent fail-closed. This job pins
`runs-on: ubuntu-latest`, which supplies gawk, so it is not a problem here.
Tracked upstream as [Morrison-Lab/gha#448](https://github.com/Morrison-Lab/gha/issues/448),
filed independently earlier today; this PR makes serodynamics a second
consumer of the affected matcher.

**The branching**, extracted from the `run:` block and driven with a stubbed
`gh` -- 11 cases covering both trigger spellings, `/reviewer` (must not fire),
an agent task, a prose mention, an ordinary comment, and both matchers
degraded. All pass.

That table was mutation-tested rather than assumed: deleting the mention branch
flips exactly the three mention cases from `dispatch` to `reply` and leaves the
other eight unchanged. The first mutation attempt was a silent no-op (the
extracted block scalar is dedented, so the anchor missed), which is why the
second run asserts the file actually changed before scoring the result.

**Diff-scoped checks**, run after committing against `origin/main...HEAD`:
0 banned glyphs in 155 added lines examined, and `check-new-line-breaks`
reports no lines missing semantic breaks.

`DESCRIPTION` is bumped to `0.1.0.9014` for `version-check`, and `NEWS.md` has
an entry. Two words were reworded out of the NEWS entry (`regex`,
`blockquote`) rather than added to `inst/WORDLIST`.

### Corrections to this body

- An earlier revision cited the mawk defect as `Morrison-Lab/gha#450`. That
  number was predicted before the issue was looked up and never existed; the
  dupe check then found the defect already tracked as
  [#448](https://github.com/Morrison-Lab/gha/issues/448). Corrected above.

## How to confirm after merge

Comment `@claude review` on any open PR. It should now acknowledge and dispatch
rather than skipping. The path needs this file on the default branch first, so
it cannot be tested from this PR's own branch.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kt6252bg9mCh6onaBQ58hY)_

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,11 +1,24 @@
 # On-demand Claude Code review of pull requests -- caller stub for the central
 # Morrison-Lab/gha reusable workflow.
 #
-# Reviews run ONLY on request: comment `/review` on a PR. Automatic review on
-# PR activity is disabled (see #266), and so is the @claude agent
-# (claude.yml), so this `/review` comment path is the only way a review runs.
-# It is a slash command rather than an `@claude ...` mention on purpose: the
-# mention path lives in claude.yml, which is switched off.
+# Reviews run ONLY on request: comment either `/review` or `@claude review` on
+# a PR. Automatic review on PR activity is disabled (see #266), and so is the
+# @claude agent (claude.yml), so this comment path is the only way a review
+# runs.
+#
+# Both spellings are accepted because only one of them was, and it was the one
+# nobody types. #282 moved review dispatch here behind `/review` alone,
+# reasoning that the mention form belonged to claude.yml -- but claude.yml had
+# just been switched off, so `@claude review` stopped doing anything at all.
+# Every one of the three review requests made in the following twelve days used
+# the mention form and was silently skipped, and `/review` was never typed once
+# (see #285).
+#
+# !! If you re-enable claude.yml, DELETE the mention path below. !!
+# The reusable claude.yml runs its own review matcher, so an `@claude review`
+# would then match both it and this job and dispatch two paid review runs --
+# against possibly different heads, since neither waits for the other. That is
+# #277, which is why upstream's stub is `/review`-only.
 #
 # The workflow_dispatch path is what dispatch-on-comment re-enters, and is
 # also how claude.yml re-dispatches a review after an @claude run pushes
@@ -37,16 +50,33 @@ on:
         type: string
 
 jobs:
-  # `/review` at the start of a PR comment (from an OWNER/MEMBER/COLLABORATOR)
-  # dispatches an on-demand review of that PR. This re-enters the
-  # workflow_dispatch path below via `gh workflow run` (the reusable
-  # workflow's jobs skip a raw issue_comment run), so it needs this file on
-  # the default branch first.
+  # A `/review` command, or an `@claude review` mention, from an
+  # OWNER/MEMBER/COLLABORATOR dispatches an on-demand review of that PR. This
+  # re-enters the workflow_dispatch path below via `gh workflow run` (the
+  # reusable workflow's jobs skip a raw issue_comment run), so it needs this
+  # file on the default branch first.
+  #
+  # The `if:` is only a cheap pre-filter -- a GitHub Actions expression cannot
+  # tell `@claude review` from `@claude, please fix the tests`. It admits any
+  # `@claude` mention and lets the steps decide, so the job also gets to REPLY
+  # to a mention it declines, rather than leaving the commenter with the same
+  # silence #285 was about.
+  #
+  # Bot comments are excluded outright. The author_association allowlist below
+  # already covers them (github-actions[bot] posts as CONTRIBUTOR here), but
+  # this job's own replies quote `@claude review` back to the thread, so a
+  # single hole in that allowlist would be a comment loop rather than a stray
+  # run. Two independent guards is cheap insurance against an expensive
+  # failure.
   dispatch-on-comment:
     if: >-
       github.event_name == 'issue_comment' &&
       github.event.issue.pull_request &&
-      startsWith(github.event.comment.body, '/review') &&
+      github.event.comment.user.type != 'Bot' &&
+      (
+        startsWith(github.event.comment.body, '/review') ||
+        contains(github.event.comment.body, '@claude')
+      ) &&
       contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
     runs-on: ubuntu-latest
     permissions:
@@ -59,6 +89,40 @@ jobs:
         with:
           workflow-ref: ${{ github.workflow_ref }}
 
+      # Decides whether an `@claude ...` mention is a REVIEW request as opposed
+      # to an agent task ("@claude, please fix the failing test"). Uses gha's
+      # canonical matcher rather than a regex of our own: it already knows the
+      # accepted phrasings (`@claude review`, `@claude, please review`,
+      # `@claude review this`), and it strips code spans and blockquotes first,
+      # so a comment that merely quotes the trigger while writing about it does
+      # not fire one.
+      #
+      # `continue-on-error` so a matcher failure degrades to "not a review"
+      # rather than reddening the job; the `/review` command path below does
+      # not depend on this step at all.
+      - name: Is this comment a review request?
+        id: review-request
+        continue-on-error: true
+        uses: Morrison-Lab/gha/.github/actions/detect-review-request@v2
+        with:
+          comment-body: ${{ github.event.comment.body }}
+
+      # Only reached when the body mentions @claude and is NOT a review request.
+      # Distinguishes a mention genuinely addressed to the bot from one quoted
+      # in prose or a code block ("as @claude said earlier"), so the
+      # agent-is-disabled reply below doesn't fire on every passing reference.
+      # Its bias is deliberately the opposite of the review matcher's: a missed
+      # reply is worse than a redundant one, so an empty result -- the step
+      # errored -- falls back to the raw `contains()` verdict the job gate
+      # already made, and replies.
+      - name: Is this a mention actually addressed to the bot?
+        id: bot-mention
+        continue-on-error: true
+        if: steps.review-request.outputs.match != 'true'
+        uses: Morrison-Lab/gha/.github/actions/detect-bot-mention@v2
+        with:
+          comment-body: ${{ github.event.comment.body }}
+
       - name: Dispatch a review for the commented PR
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


d-morrison:
## No Claude review will run on this PR, and that is expected

Flagging it up front so nobody waits on a verdict that cannot arrive -- which
is, with some irony, the same failure mode this PR exists to fix.

The reusable review workflow carries a self-modification guard. It derives the
**caller's** workflow path from `github.workflow_ref` and skips every review
step when the PR's changed files include that path
(`Morrison-Lab/gha/.github/workflows/claude-code-review.yml@v2`, the
`selfmod` step). In this repo the caller is
`.github/workflows/claude-code-review.yml` -- exactly the file this PR edits.
So `self_mod=true`, and the review is skipped by design: the action's OIDC
token exchange validates the calling workflow's content against the default
branch, which cannot match until this merges.

Two consequences:

- The guard keys on this PR's own diff, so **re-triggering will not clear it**.
  Not worth spending runs on.
- `v2` is new enough to carry the fix that surfaces this as a gray *skipped*
  `require-review` plus an explanatory comment, rather than a green check over
  an absent review. If you see that comment, it is this, not a failure.

Copilot has been requested as the external reviewer instead.

## Self-review

Posting one because no bot verdict is available, held to the standard the bot
would apply rather than a structural skim.

**Correctness of the trigger logic.** Both matchers are `continue-on-error`,
so I checked each degraded path rather than only the happy one. With
`steps.review-request.outputs.match` empty, `MENTION_IS_REVIEW` is an empty
string -- defined, so `set -u` does not fire -- and the `/review` branch is
untouched, so a matcher outage cannot break the command path. With
`steps.bot-mention.outputs.match` empty, the reply is *sent* rather than
withheld, matching that action's documented bias (a missed reply costs more
than a redundant one). Both are covered in the offline table.

**The loop question.** This job now writes `@claude review` into the thread in
two places -- the acknowledgement and the agent-is-off reply. That text
re-enters the pre-filter, so I traced it: `github-actions[bot]` posts with
`author_association: CONTRIBUTOR`, which the allowlist excludes, and the new
`user.type != 'Bot'` condition excludes it again independently. Belt and
braces, deliberately, because the failure mode is unbounded comment spam
rather than one stray run.

**What I could not verify, stated plainly.** The end-to-end path cannot be
exercised from this branch: `gh workflow run` resolves the workflow file from
the **default branch**, so the new gate only takes effect after merge. The
matcher and the branching were verified offline instead (details in the PR
body), and the post-merge confirmation is one comment on any open PR.

**One thing reviewers should push back on if they disagree.** The
agent-is-off reply is a judgment call, not a bug fix. It adds a bot comment to
every genuine `@claude` mention that is not a review request. I think that is
right -- those comments are asking for work that will never happen, and
silence is what let this bug survive twelve days -- but it is the part of this
PR most reasonably contested, and it can be dropped without affecting the
trigger fix.

---

d-morrison:
## `docs-check` is red here, and it is not this PR

Diagnosed rather than waved off, and not fixed here on purpose. Tracked as
#288.

**What it reports:** `roxygenise()` rewrites `DESCRIPTION` and `NAMESPACE`, so
the check's "no diff after documenting" assertion fails.

**Why:** the workflow installs `any::roxygen2`, which is now **8.1.0**;
`DESCRIPTION` commits `Config/roxygen2/version: 8.0.0`. The version field gets
rewritten and 8.1.0 regenerates `NAMESPACE` differently. Purely an upstream
release, nothing in the repo.

**Why it is not mine, established with a control rather than asserted:**

- #283 is a `.ai-config` submodule bump with no R files at all, and its
  `docs-check` run at `07:22:30Z` fails identically -- five minutes before this
  PR existed.
- `Config/roxygen2/version: 8.0.0` is byte-identical on `main` and this branch.
- This diff touches `DESCRIPTION` only at the `Version:` line, and touches no
  `R/`, `man/`, or `NAMESPACE` file. `git diff origin/main...HEAD -- NAMESPACE
  man/ R/` returns nothing.
- `docs-check` was green on `main` at 8b53be4 (2026-07-31), which is this PR's
  base commit. The drift arrived between then and now.

**Why I am not fixing it here.** The fix is to re-document under 8.1.0 and
commit the regenerated `DESCRIPTION` and `NAMESPACE` -- a repo-wide change
unrelated to a review trigger, which belongs in its own PR where a reviewer can
see the regenerated diff on its own. It also needs roxygen2 **8.1.0**
specifically; this sandbox has no roxygen2 installed, so anything I generated
here would risk a different version's output and leave the check red for a
second reason.

Everything else on this PR is green, including the two checks the change is
actually responsible for -- `version-check` and `Check Changelog Action` -- plus
`Spellcheck` and `lint-changed-files`.

One knock-on worth knowing: #288 will land on `main` before or after this, and
whichever goes second will want a `main` merge to pick up the regenerated
files. No conflict either way -- the two diffs touch disjoint files apart from
`DESCRIPTION`, and on different lines.

---

d-morrison:
Independent post-merge verification of this change, since it landed without a
substantive external review: Copilot refused on quota, and the Claude review
was skipped by the reusable workflow's self-modification guard (`v2` lines
391-406) because the diff edits `claude-code-review.yml` itself. That guard is
correct, but it means this fix reached `main` unreviewed by anything.

**Result: no defects found.** The gate behaves correctly on every case I could
construct from real data.

## Method

I extracted `detect-review-request.sh`, `detect-bot-mention.sh` and
`strip-non-invoking-markup.sh` from `Morrison-Lab/gha` at the `v2` tag and
simulated this job end to end -- the job-level `if:`, both matcher steps with
their real `if:`/`continue-on-error` semantics, and the shell's `TRIGGER`
branch -- against 20 comment bodies, 16 of them **copied verbatim from this
repo's PR #230**.

## What the merged gate does

| real comment from #230 | result |
| --- | --- |
| the 3 member review requests that got silence (07-31, 08-11, 08-12) | dispatches |
| the polite `please review` form (#5030118447) | dispatches |
| 5 real agent tasks (`please update this PR`, `please make the following modifications`, ...) | replies "agent is off" |
| bot post-step comment with the mention in a code span | job skipped (Bot author) |
| a **blockquoted** review request (quote-reply, #5040690360) | silent skip |
| the `@copilot` task (#5109374176) | job skipped |
| mention inside a fenced block | silent skip |
| `@claude reviewer` | replies, does not dispatch |
| non-member, either spelling | job skipped |

The two I paid closest attention to, because they are the ways this kind of
change usually goes wrong:

- **The acknowledgement comment cannot re-trigger the job.** It carries the
  trigger phrase, but in a code span, and the matcher strips code spans before
  matching. Belt and braces with the `Bot` author filter and the fact that
  GitHub raises no events for comments posted with `GITHUB_TOKEN`.
- **The quote-reply case.** A real comment on #230 quotes a review request
  inside a blockquote; it correctly neither dispatches nor replies.

## One note, not a defect

The `Bot` filter is doing real work here: three of the four bodies that would
otherwise have drawn an "agent is off" reply are the bot's own comments. Worth
keeping in mind if that filter is ever relaxed.

Separately, `Morrison-Lab/gha#451` records a portability bug in the shared
matcher (it returns `false` for every body under `mawk`). **This workflow is
not exposed** -- `claude-code-review.yml@v2` hard-codes `runs-on: ubuntu-latest`,
whose `awk` is gawk. Noting it only so the link exists if that ever changes.

I had an in-flight PR for the same issue (#287); this one is a superset of it,
so I have closed mine.


---

PR #287, opened 2026-08-12
ci: accept an `@claude review` mention alongside `/review`

### LINKED ISSUE #285: ci: `@claude review` is silently ignored -- the only review trigger is `/review`, which nobody uses

d-morrison:
Closes #285.

Also resolves `Morrison-Lab/gha#450`, the same bug filed in the `gha` repo
because the session that wrote it could not reach this one. The fix belongs
here: `examples/` in `gha` is copy-and-edit, so landing the upstream variant
(`Morrison-Lab/gha#449`) would not change this repo.

## Problem

Disabling the agent (#282) left `/review` as the only phrasing that dispatched
a review. The mention is what people actually type, and it produced a run with
every job skipped and no comment posted, which is indistinguishable from a
broken bot. Three review requests from members went unanswered that way
(#285 has the table).

## Change

`dispatch-on-comment` now accepts either form. The job-level `if:` stays a
cheap pre-filter and the authoritative decision moves into the steps, where a
real matcher can see Markdown and word boundaries:

- `/review` at the start of a comment, as before, or
- a mention that `Morrison-Lab/gha`'s own `detect-review-request@v2` action
  classifies as a review request.

Reusing that action rather than writing a second inline pattern is the point:
this path inherits its code-span and fenced-block stripping and its
accepted-phrasing set, and cannot drift away from them the way two copies of
the pattern did in #277.

The mention is accepted **only because `claude.yml` is off**. `claude.yml`'s
re-enable instructions now say to delete this path in the same change, since
answering one comment in both workflows dispatches two paid reviews against
possibly different heads.

## Verification

I ran the real matcher (extracted from `gha` at the `v2` tag) against 26 cases,
20 of them **real comment bodies taken from this repo's own PR #230**, through
a simulation of the full gate (job `if:` -> matcher step -> shell `TRIGGER`
logic). 26 examined, 0 failures.

| comment | before | after |
| --- | --- | --- |
| `/review`, `/review please` | dispatches | dispatches |
| `/reviewer` | no dispatch | no dispatch |
| the 3 real member review requests that got silence | **silence** | **dispatches** |
| real agent tasks (`@claude, please update this PR ...`) | no dispatch | no dispatch |
| the `@claude` mention quoted in a code span, in a real bot comment | no dispatch | no dispatch |
| a real **blockquoted** `@claude, please review` (quote-reply) | no dispatch | no dispatch |
| a mention asking a question | no dispatch | no dispatch |
| non-member comment, either form | job skipped | job skipped |

Two loop checks, since the mention path is new:

- The acknowledgement this job posts contains `` `@claude review` `` in a code
  span. Fed back through the matcher it returns `false`, so it cannot
  re-trigger itself even setting aside that GitHub raises no events for
  comments posted with `GITHUB_TOKEN`. Both are noted in the code.
- `@claude reviewer` still does not match.

Both workflow files parse as YAML. No banned punctuation in 107 added lines,
and `check-new-line-breaks` reports no missing semantic breaks.

## What this PR cannot verify before merge

- **The `issue_comment` path itself.** GitHub runs the **default branch's**
  copy of a workflow for `issue_comment` events, so commenting on this PR
  exercises `main`'s version, not this one. The file's own comment already
  notes it "needs this file on the default branch first." Verified by
  simulation instead, above.
- **A Claude review of this PR.** The diff edits
  `.github/workflows/claude-code-review.yml`, which trips the reusable
  workflow's self-modification guard (`v2`, lines 391-406): it sets
  `self_mod=true` and skips every review step, because the action's OIDC
  workflow-validation 401s until the file matches the default branch. So
  `review / require-review` will show a **gray skip, not an all-clear** --
  please do not read it as one. Copilot is requested as the external reviewer,
  and a self-review is posted below.

## Out of scope

- `docs-check` is already failing on `main`: it failed on this PR's empty
  scaffold commit, whose tree is byte-identical to `main`'s
  (`1a34593bad0b4b43d1cbab39d2edad3ae9520a59`). `devtools::document()` dirties
  `DESCRIPTION` and `NAMESPACE` there. Not caused by this change; filed
  separately.
- Replying to a trusted `@claude` mention that is *not* a review request, so
  "the agent is off" is visible rather than inferred, is the second half of
  #285. Deferred and filed separately rather than folded in here: which
  mentions deserve an automated reply is a judgment call worth its own
  review, and a reply on every prose mention would be noise.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -52,18 +52,24 @@ jobs:
   # `gh workflow run` (the reusable workflow's jobs skip a raw issue_comment
   # run), so it needs this file on the default branch first.
   #
-  # Both clauses below are cheap PRE-FILTERS only, kept broad on purpose: the
-  # authoritative match happens in the steps below, where a real matcher can
-  # see Markdown and word boundaries. Keeping the gate here loose and the
-  # decision there is what lets one matcher serve both this workflow and
-  # claude.yml.
+  # The clauses below are cheap PRE-FILTERS only: the authoritative match
+  # happens in the steps, where a real matcher can see Markdown and word
+  # boundaries. Keeping the gate loose and the decision there is what lets one
+  # matcher serve both this workflow and claude.yml.
+  #
+  # The mention pre-filter requires the word `review` as well as the mention.
+  # That cannot drop a request the matcher would have accepted -- its pattern
+  # ends in `(re-?)?review`, so every body it matches contains `review`
+  # literally -- and it keeps a runner from starting for the commonest kind of
+  # mention by far, a task for the agent that never asks for a review.
   dispatch-on-comment:
     if: >-
       github.event_name == 'issue_comment' &&
       github.event.issue.pull_request &&
       contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association) &&
       (startsWith(github.event.comment.body, '/review') ||
-      contains(github.event.comment.body, '@claude'))
+      (contains(github.event.comment.body, '@claude') &&
+      contains(github.event.comment.body, 'review')))
     runs-on: ubuntu-latest
     permissions:
       actions: write        # dispatch this workflow via `gh workflow run`
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -1,11 +1,20 @@
 # On-demand Claude Code review of pull requests -- caller stub for the central
 # Morrison-Lab/gha reusable workflow.
 #
-# Reviews run ONLY on request: comment `/review` on a PR. Automatic review on
-# PR activity is disabled (see #266), and so is the @claude agent
-# (claude.yml), so this `/review` comment path is the only way a review runs.
-# It is a slash command rather than an `@claude ...` mention on purpose: the
-# mention path lives in claude.yml, which is switched off.
+# Reviews run ONLY on request. Two comment triggers are accepted, both by the
+# dispatch-on-comment job below:
+#   * `/review` at the start of a comment, and
+#   * an `@claude review` mention.
+# Automatic review on PR activity is disabled (see #266), and so is the
+# @claude agent (claude.yml), so these two comment paths are the only ways a
+# review runs.
+#
+# The mention form is accepted HERE only because claude.yml is switched off.
+# Normally an `@claude` mention wakes that workflow, which routes a review
+# request to this one; answering it in both places dispatches two paid reviews
+# for a single comment, against possibly different heads (see #277). So if the
+# agent is ever re-enabled, the `@claude` clause in dispatch-on-comment's `if:`
+# and the mention-matcher step must be deleted at the same time.
 #
 # The workflow_dispatch path is what dispatch-on-comment re-enters, and is
 # also how claude.yml re-dispatches a review after an @claude run pushes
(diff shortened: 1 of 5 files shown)


d-morrison:
Closing: superseded by #286, which merged while this was in review.

#286 is a strict superset of this PR. It makes the same change (accept a
mention alongside `/review`, matched with `Morrison-Lab/gha`'s canonical
`detect-review-request` action rather than a second inline pattern), updates
`claude.yml`'s re-enable instructions with the same warning, and additionally:

- composes `detect-bot-mention` and **replies** when a trusted mention is not a
  review request, which is the second half of #285 that I had deferred to #290
  (now closed as already implemented), and
- filters `github.event.comment.user.type != 'Bot'`.

It also bumped `DESCRIPTION` to the same `0.1.0.9014` this branch used, so the
branch is now at version parity with `main` and `version-check` would fail here
regardless. Resolving toward `main` would leave essentially nothing:
`inst/WORDLIST` gaining `blockquote`, which is needed only because *my*
`NEWS.md` wording used the word.

One thing here was actively **wrong** against what merged, and is worth
recording rather than quietly dropping. I had narrowed the mention pre-filter
to require the word `review` as well as the mention, to avoid starting a runner
for agent-task mentions. That is safe against this PR's behaviour, but it would
have **broken #286's reply path**: that path exists precisely to answer
mentions that are *not* review requests, and those are exactly the bodies the
narrowing skips before the job starts. A good argument for the superset landing
instead of this one.

The verification work is not wasted -- I re-pointed it at the merged
implementation and posted the results on #286. Summary: no defects found across
20 bodies, 16 of them copied verbatim from #230.

Issues opened along the way that stay open, since neither is addressed by #286:

- #289 -- `docs-check` is failing on `main` (`devtools::document()` dirties
  `DESCRIPTION` and `NAMESPACE`); proven pre-existing via a byte-identical
  tree, not caused by either PR.
- `Morrison-Lab/gha#451` -- the shared matcher silently returns `false` for
  every body under `mawk`. Not exposed here (`claude-code-review.yml@v2` pins
  `ubuntu-latest`).


---

PR #291, opened 2026-08-12
docs: regenerate NAMESPACE and DESCRIPTION under roxygen2 8.1.0

### LINKED ISSUE #288: ci: `docs-check` fails on every PR -- roxygen2 8.1.0 rewrites DESCRIPTION and NAMESPACE against the pinned 8.0.0
## Symptom

`docs-check` (`.github/workflows/R-check-docs.yml`) fails on **every** open PR,
including ones that touch no R code at all. It reports:

```
ℹ Setting Config/roxygen2/version to "8.1.0"
Writing 'NAMESPACE'
...
Changes found after documenting.
DESCRIPTION
NAMESPACE
Please update documentation using the devtools::document() command in R.
```

## Cause

The workflow installs `any::roxygen2`, i.e. whatever is current on RSPM. That
is now **8.1.0**. `DESCRIPTION` commits `Config/roxygen2/version: 8.0.0`, so
`roxygenise()` rewrites that field, and 8.1.0 also regenerates `NAMESPACE`
differently from 8.0.0. Both show up as an uncommitted diff, which is exactly
what the check is built to fail on.

Nothing in the repository changed to cause this. `docs-check` was green on
`main` as recently as 8b53be4 (2026-07-31); roxygen2 8.1.0 landed after that,
and the next PR to run the check inherited the failure.

## It is not any one PR's fault

Confirmed with a control rather than inferred. #283 is a `.ai-config`
submodule bump -- no R files, no `man/`, no `NAMESPACE` -- and its run at
`2026-08-12T07:22:30Z`
([94040969603](https://github.com/UCD-SERG/serodynamics/actions/runs/31361755505/job/94040969603))
fails identically, five minutes before #286 existed. #286
([94047470322](https://github.com/UCD-SERG/serodynamics/actions/runs/31575787157/job/94047470322))
fails the same way and touches only workflow YAML, `NEWS.md`, and the
`Version:` line of `DESCRIPTION`.

This also explains the report on #230 from 2026-08-05: *"the failing
documentation check as document() does not show any changes in the local
environment."* A contributor whose local roxygen2 is 8.0.0 genuinely sees no
diff -- the drift only exists against CI's newer copy.

## Fix

Re-document under 8.1.0 and commit the result, in a PR of its own:

```r
roxygen2::roxygenise()   # with roxygen2 8.1.0 installed
```

Expect `DESCRIPTION` (`Config/roxygen2/version`) and `NAMESPACE` to change; a
reviewer should confirm nothing else does. Whoever does it needs 8.1.0
specifically -- running an older roxygen2 will produce a different diff and
leave the check red.

## Worth deciding at the same time

`any::roxygen2` means the next roxygen2 release does this again, to every open
PR at once, with no change on our side. Two options:

- **Pin** the version in `R-check-docs.yml` and bump it deliberately. Turns a
  surprise into a scheduled chore, at the cost of drifting from the tool
  contributors install by default.
- **Keep it floating** and accept the periodic re-document. Cheaper most of the
  time, but the failure always lands on whoever happens to open the next PR,
  and reads as their fault.

I have no strong preference; flagging it because the current setup makes an
unrelated contributor pay for an upstream release.

d-morrison:
Closes #288. Unblocks `docs-check` on every open PR.

## What was wrong

`docs-check` installs `any::roxygen2`, so it runs whatever is current. That is
now **8.1.0**, and the committed files were generated with 8.0.0 -- so
`roxygenise()` rewrites both, and the check's "no diff after documenting"
assertion fails.

It failed on every PR, including ones with no R code: #283 (a submodule bump)
and #286 (workflow YAML) both failed identically, and #286's base commit
8b53be4 had been green on `main` twelve days earlier. Nothing in the repository
changed; roxygen2 released.

It also explains the report on #230 from 2026-08-05 -- *"document() does not
show any changes in the local environment"*. A contributor on 8.0.0 genuinely
sees no diff; the drift exists only against CI's newer copy.

## The change

- **`NAMESPACE`** -- formatting only. Where several symbols come from one
  package, 8.1.0 groups them into a single `importFrom()` call instead of a
  line each.
- **`DESCRIPTION`** -- `Config/roxygen2/version` goes to 8.1.0, and
  `RoxygenNote` is dropped, which 8.1.0 no longer writes.

Per your call on #288, the roxygen2 version **stays floating** rather than
pinned. A later release will need the same treatment; that is the accepted
trade.

## Verification

**Generated with CI's own version**, not merely a recent one: roxygen2 8.1.0,
on Ubuntu noble, installed from the same
`packagemanager.posit.co/cran/__linux__/noble/latest` binary repo the workflow
uses.

**Matching the environment turned out to matter**, and is worth recording. A
first run produced a *third* changed file, `man/expect_snapshot_data.Rd`,
because `testthat` sits in `Suggests` and was not installed, so that file's
`@inheritDotParams testthat::expect_snapshot_file` could not resolve and
roxygen emitted the topic without the inherited arguments. roxygen said so
outright rather than degrading quietly:

```
x expect_snapshot_data.R:10: @returns refers to un-installed package testthat.
x In topic 'expect_snapshot_data': @inheritDotParams failed because testthat is not installed.
```

Those warnings, plus the extra file being one CI never reports, are what
flagged the environment as wrong rather than the output. Installing the full
`Suggests` set reproduced CI's exact two-file result with no warnings.

**The NAMESPACE change is semantically inert**, established by parsing rather
than by reading the diff. `parseNamespaceFile()` on both versions gives
identical sets:

| | before | after |
| --- | --- | --- |
| exports | 22 | 22 |
| imports | 27 | 27 |
| S3 methods | 1 | 1 |

That comparator was itself checked against a deliberately dropped import
(`stats::quantile`), which it reports by name -- so "identical" is a measured
result rather than a checker that cannot fail.

**CI's assertion reproduced locally.** Re-running `roxygenise()` against the
committed tree and applying the workflow's own
`git diff-index HEAD --name-only -- man/ NAMESPACE DESCRIPTION` returns empty.

**Other checks:** `spelling::spell_check_package()` is clean (0 misspellings --
the real package, not a proxy). 0 banned glyphs in 52 added lines examined, and
`check-new-line-breaks` reports none missing. `DESCRIPTION` is bumped to
`0.1.0.9015`.

## Corrections to this body

Both found in review of the corpus entry this session wrote from the same
material ([Morrison-Lab/ai-config#1401](https://github.com/Morrison-Lab/ai-config/pull/1401)),
and corrected here so the two do not disagree. Neither affects the diff.

- An earlier revision said roxygen **silently** emitted a degraded topic. It
  did not -- it printed the two warnings quoted above, naming the tag and the
  missing package. Those warnings were in fact the first signal acted on, so
  calling the extra file "the tell" understated a better one.
- An earlier revision attributed that file's degradation to `testthat` **and**
  `rjags`. Only `testthat` is causally connected to it: `R/expect_snapshot_data.R`
  contains no `rjags` reference. `rjags` was absent too and was flagged on
  `R/run_serodynamics.R`, whose output did not change.

## Note

I did not touch renv -- this repository has no `renv.lock`, no `renv/`, and no
renv reference in any workflow. Its `.Rprofile` uses `pacman` for interactive
sessions only. Nothing to update.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: serodynamics
 Title: Modeling Longitudinal Antibody Responses to Infection
-Version: 0.1.0.9014
+Version: 0.1.0.9015
 Authors@R: c(
     person("Peter", "Teunis", , "p.teunis@emory.edu", role = c("aut", "cph"),
            comment = c("Author of the method and original code.",
@@ -78,5 +78,4 @@ Depends:
     R (>= 4.1.0)
 LazyData: true
 VignetteBuilder: knitr, quarto
-Config/roxygen2/version: 8.0.0
-RoxygenNote: 8.0.0
+Config/roxygen2/version: 8.1.0
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -23,30 +23,40 @@ export(serodynamics_example)
 export(sim_case_data)
 export(sim_n_obs)
 export(summarize_posterior)
-importFrom(dplyr,.data)
-importFrom(dplyr,all_of)
-importFrom(dplyr,filter)
-importFrom(dplyr,mutate)
-importFrom(dplyr,n)
-importFrom(dplyr,reframe)
-importFrom(dplyr,rename)
-importFrom(dplyr,rowwise)
-importFrom(dplyr,select)
-importFrom(dplyr,ungroup)
+importFrom(dplyr,
+  .data,
+  all_of,
+  filter,
+  mutate,
+  n,
+  reframe,
+  rename,
+  rowwise,
+  select,
+  ungroup
+)
 importFrom(ggmcmc,ggs)
-importFrom(ggplot2,autoplot)
-importFrom(ggplot2,theme_bw)
-importFrom(ggplot2,vars)
-importFrom(rlang,.data)
-importFrom(rlang,.env)
+importFrom(ggplot2,
+  autoplot,
+  theme_bw,
+  vars
+)
+importFrom(rlang,
+  .data,
+  .env
+)
 importFrom(runjags,run.jags)
-importFrom(serocalculator,get_biomarker_levels)
-importFrom(serocalculator,get_biomarker_names)
-importFrom(serocalculator,get_biomarker_names_var)
-importFrom(serocalculator,get_values)
-importFrom(serocalculator,ids)
-importFrom(serocalculator,ids_varname)
-importFrom(stats,complete.cases)
-importFrom(stats,quantile)
+importFrom(serocalculator,
+  get_biomarker_levels,
+  get_biomarker_names,
+  get_biomarker_names_var,
+  get_values,
+  ids,
+  ids_varname
+)
+importFrom(stats,
+  complete.cases,
+  quantile
+)
 importFrom(tidyr,pivot_wider)
 importFrom(utils,read.csv)
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,21 @@
 
 ## Internal
 
+* Regenerated `NAMESPACE` and `DESCRIPTION` under roxygen2 8.1.0 (#288).
+  The documentation check installs whatever roxygen2 is current rather than a
+  fixed version, so the 8.1.0 release started rewriting both files against the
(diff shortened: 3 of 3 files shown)


d-morrison:
@claude review

---

